PR #17, opened 2019-06-03
Setup testing with Travis

marcelm:
This is what Travis is configured to do:

- Install Miniconda
- Create a Conda environment with all the required dependencies
- Download the test dataset
- Run BLR

The `.travis.yml` file is somewhat inspired by the one [in the Conda documentation](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/use-conda-with-travis-ci.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,31 @@
+language: python
+
+dist: xenial
+
+cache:
+  directories:
+    - $HOME/.cache/pip
+
+python:
+  - "3.6"
+
+before_install:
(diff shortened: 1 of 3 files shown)


marcelm:
The test actually fails at the moment, but I would like to work on that problem in a separate issue/PR.

Usually, one would not merge a PR that makes CI fail, but in this case, since we’re just getting started, I’ll ask you to do it anyway.

marcelm:
I’ve pushed another commit that makes it easier to run the tests on the commandline by running `tests/run.sh` (you need to first download the test data and setup the `paths.txt` manually).

PR #19, opened 2019-06-03
Always keep log files

marcelm:
Log files make it possible to find out what has gone wrong when there is a
problem, and they help reproducibility, for example, because they contain
the versions of programs that were run.

Since log files are small, space savings cannot be an argument.

If clutter is a concern, log files could be kept in a separate log/ directory,
but that would be a different change.

Also, note that the variable is always misspelled as `keep_logiles` (without `f`), except where it is being set to `true`, so this feature did not really work anyway and would always delete log files (with or without `-k`). (Except for `cpicard.log`, which contains an extra `c` as a typo and is therefore always kept.)

### DIFF
--- a/BLR_automation.sh
+++ b/BLR_automation.sh
@@ -37,7 +37,6 @@ processors=1
 mailing=false
 remove=false
 duplicate_rmdup=false
-keep_logiles=false
 heap_space=90
 index_nucleotides=3
 threshold=0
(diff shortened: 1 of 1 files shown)


FrickTobias:
Let's keep them in a separate log directory but, as you suggest, implement them in a separate commit.

PR #27, opened 2019-06-05
No need to derive from object explicitly

marcelm:
In Python 3, every class implicitly derives from object.

In Python 2, there was a difference: Leaving out `(object)` would give you an
old-style class and including it a new-style class. Python 3 only has
new-style classes.


FrickTobias:
The more you know!

PR #38, opened 2019-06-10
Move some more steps to the Snakefile

marcelm:
All adapter trimming steps and the barcode extraction step are now in the Snakefile, and the automation script has gotten smaller correspondingly.


marcelm:
`threads:` gives the *maximum* number of threads that the rule can reasonably use. When you run snakemake, you then provide the option `--cores` (or `-j`) to give it the actual number of threads to use. Snakemake then fills the `{threads}` with the actual value.

In other words: Although it says `threads: 20` in the rule, if you run snakemake with `-j 8`, `{threads}` will be only be 8.

PR #49, opened 2019-06-25
Remove pythonversion

pontushojer:
Remove function pythonversion from utils.py. This was previously used to confirm that python3 was used to run the script but is no longer relevant as this is now specified within in the setup. Also removed force_run argument from multiple script which was used as an override for the version-check.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -2,8 +2,8 @@
 rule trim_r1_handle:
     "Trim away E handle on R1 5'. Also removes reads shorter than 85 bp."
     output:
-        r1_fastq=temp("{dir}/trimmed-a.1.fastq.gz"),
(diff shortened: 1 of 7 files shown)


marcelm:
Since you don’t have write access to Pontus’ branch, you cannot easily change this PR yourself. The easiest way is to merge first and then just make the change yourself.

pontushojer:
I changed the wording on my branch. 

> Great work!
> 
> Just one minor thing, I see the clusterrmdup documentation now states `Run analysis even if not no barcode if found.`. I think it's supposed to be `Run analysis even if barcode tags are not found for all reads.`.
> 
> I'll merge after you make the change, or is it some way I can do it while merging? For such a simple thing as this I'm thinking it could be simpler.

I edited the wording now anyway

Issue #43, opened 2019-06-17
filter_clusters Cleanup


FrickTobias:
filter clusters is written a long time ago and has a lot of clumsy code which can be significantly improved & simplified in order to make it more readable.

PR #52, opened 2019-07-01
Filtercluster rewrite

FrickTobias:
Resolving #43 and rewriting filtercluster to be more readable while keeping the same function.

Briefly what was done:
- Move everything regarding counting of reads/molecules in barcodes to their own objects
- Move most main function code to smaller functions 
- Change variable names to more intuitive names
- Add & update documentation
- Write (more) according to pep8 (Glad to accept feedback on this point in particular)
- Made standard functionality to be writing filtered bam files with optional molecule/barcode stats generation instead of the other way around
- Removed outdated functions & options
- Changed statistics output format to be one file for molecule stats and one for barcodes

**Notable changes**
- Remove utils dependency
- tqdm for reporting over iterations (external module)
- Changed args (option names & positional args)

There is **one change** to output, the filtered reads have a slight format change. Prev format: 

```<header>_<bc_seq>_<bc_clust_id>``` => ```<header>_<bc_seq>```

New format:

```<header>_<bc_seq>_<bc_clust_id>``` => ```<header>_FILTERED-<bc_seq>_<bc_clust_id>```

Where ```bc_clust_id``` corresponds to the error-corrected barcode.

### DIFF
--- a/src/blr/cli/filterclusters.py
+++ b/src/blr/cli/filterclusters.py
@@ -14,7 +14,6 @@ logger = logging.getLogger(__name__)
 
 def main(args):
     summary = Summary()
-
     logger.info(f'Running analysis with {"{:,}".format(args.window)} bp window size')
 
     # Build allMolecules.final_dict[barcode][moleculeID] = molecule
@@ -33,22 +32,22 @@ def main(args):
         with pysam.AlignmentFile(args.output, 'wb', template=openin) as openout:
             logger.info("Writing filtered bam file")
             for read in tqdm(openin.fetch(until_eof=True)):
-                BC_id = fetch_bc(pysam_read=read, barcode_tag=args.barcode_tag)
+                barcode = fetch_bc(pysam_read=read, barcode_tag=args.barcode_tag)
 
-                # If BC_id is not in allMolecules there the barcode does not have enough proximal reads to make a single molecule
+                # If barcode is not in allMolecules there the barcode does not have enough proximal reads to make a single molecule
                 # If the barcode has too many barcodes, remove it from the read
-                if BC_id in allMolecules.final_dict and len(allMolecules.final_dict[BC_id]) > args.Max_molecules:
-                    read = strip_barcode(read=read ,barcode_tag=args.barcode_tag)
+                if barcode in allMolecules.final_dict and len(allMolecules.final_dict[barcode]) > args.Max_molecules:
+                    read = strip_barcode(pysam_read=read ,barcode_tag=args.barcode_tag)
 
                     summary.reads_with_removed_barcode += 1
-                    if not BC_id in summary.barcode_removal_set:
-                        summary.barcode_removal_set.add(BC_id)
-                        summary.number_removed_molecules += len(allMolecules.final_dict[BC_id])
+                    if not barcode in summary.removed_barcodes:
+                        summary.removed_barcodes.add(barcode)
+                        summary.removed_molecules += len(allMolecules.final_dict[barcode])
 
                 openout.write(read)
 
     # Stats to output files and stdout
-    summary.printStats(barcode_tag=args.barcode_tag, threshold=args.threshold, allMolecules=allMolecules)
+    summary.printStats(barcode_tag=args.barcode_tag, allMolecules=allMolecules)
     if args.stats_file:
         logger.info("Writing statistics files")
         summary.writeMoleculeStats(output_prefix=args.stats_file, allMolecules=allMolecules)
@@ -62,8 +61,8 @@ def build_molecule_dict(pysam_openfile, barcode_tag, window, min_reads, summary)
     for read in tqdm(pysam_openfile.fetch(until_eof=True)):
 
         # Fetches barcode and genomic position. Position will be formatted so start < stop.
-        BC_id = fetch_bc(pysam_read=read, barcode_tag=barcode_tag, summary=summary)
-        if BC_id and read.is_unmapped == False:
+        barcode = fetch_bc(pysam_read=read, barcode_tag=barcode_tag, summary=summary)
+        if barcode and read.is_unmapped == False:
             read_start, read_stop = sorted((read.get_reference_positions()[0], read.get_reference_positions()[-1]))
 
             # Commit molecules between chromosomes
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by FrickTobias** [comment]
> @FrickTobias. I understand the reason to not allow overlapping reads in general (we assume it is a single molecule). With tagmentation libraries there is however a small overlap due to that the adaptor insertion that results in a 9 bp duplication, see [Reznikoff, 2003 figure 4](https://onlinelibrary.wiley.com/doi/full/10.1046/j.1365-2958.2003.03382.x). Thus for these libraries we should not remove sequences that overlap within 9 bp. I would suggest we add an option to allow overlaps of 8-10 bp.

I agree, however since the script underwent quite extensive changes I think it better to resolve that in a separate PR so it can keep the exact same functionality for now. This makes it easier to guarantee functionality has not been changed accidentally.

**Comment by FrickTobias** [comment]
The conflict comes from #49 where 
- the same statements has been removed
- An argparse option's help message has been updated upstream
so it should be fine to just remove the master version of the argparser function.

**Comment by pontushojer** [comment]
> > @FrickTobias. I understand the reason to not allow overlapping reads in general (we assume it is a single molecule). With tagmentation libraries there is however a small overlap due to that the adaptor insertion that results in a 9 bp duplication, see [Reznikoff, 2003 figure 4](https://onlinelibrary.wiley.com/doi/full/10.1046/j.1365-2958.2003.03382.x). Thus for these libraries we should not remove sequences that overlap within 9 bp. I would suggest we add an option to allow overlaps of 8-10 bp.
> 
> I agree, however since the script underwent quite extensive changes I think it better to resolve that in a separate PR so it can keep the exact same functionality for now. This makes it easier to guarantee functionality has not been changed accidentally.

Agreed. I have however made an issue (#50) that references this so we can included it later.

File: src/blr/cli/filterclusters.py [hunk 1]
```diff
@@ -1,367 +1,306 @@
...
+    defined as more than <min_reads> within <window> from each other.
+    :param pysam_openfile: Pysam open file instance.
+    :param barcode_tag: Tag used to store barcode in bam file (usually BC).
+    :param window: Max distance between reads to include in the same molecule.
+    :param min_reads: Minimum reads to include molecule in allMolecules.final_dict
+    :param summary: Custom summary intsance
+    :return: dict[barcode][molecule] = moleculeInstance
+    """
+
+    allMolecules = AllMolecules(min_reads=min_reads)
```
pontushojer: So as i understand PEP8 (You wanted comments on this ;) ) this naming of instance variables with uppercase should be avoided. I see this also for the class methods which should also be all lowercase in my understanding of the quidelines. Options in this case are: `allmolecules` or `all_molecules`, whichever is more clear. This could however be part of a later pull request.

src/blr/cli/filterclusters.py
```diff
@@ -1,367 +1,306 @@
...
+                        help="Threshold for how many reads are required for including given molecule in statistics "
+                             "(except_reads_per_molecule). DEFAULT: 4")
+    parser.add_argument("-w", "--window", metavar="<INTEGER>", type=int, default=30000,
+                        help="Window size cutoff for maximum distance in between two reads in one molecule. DEFAULT: "
+                             "30000")
+    parser.add_argument("-bc", "--barcode_tag", metavar="<STRING>", type=str, default="BC",
+                        help="Bam file tag where barcode is stored. DEFAULT: BC")
+    parser.add_argument("-s", "--stats_file", metavar="<PREFIX>", type=str,
+                        help="Write barcode/molecule statistics files. DEFAULT: None")
+    parser.add_argument("-M", "--Max_molecules", metavar="<INTEGER>", type=int, default=500,
```
pontushojer: In this case I believe the long-form of the option, i.e. `--Max_molecules` , should be written in lowercase letter even though the short-form option is `-M`. This is because it will be stored under the varible name `args.Max_molecules`which is not according to PEP8 convention.

src/blr/cli/filterclusters.py
```diff
@@ -1,367 +1,306 @@
...
+                                            min_reads=args.threshold, summary=summary)
+
+        summary.tot_reads = infile.mapped + infile.unmapped
+        summary.unmapped_reads = infile.unmapped
+        summary.mapped_reads = infile.mapped
+        summary.non_analyzed_reads = summary.unmapped_reads + summary.non_tagged_reads + summary.overlapping_reads_in_pb
+
+    # Writes filtered out
+    with pysam.AlignmentFile(args.x2_bam, "rb") as openin:
+        with pysam.AlignmentFile(args.output, "wb", template=openin) as openout:
```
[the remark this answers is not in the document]
FrickTobias: Sure, seems like a standard enough statement it wouldn't affect readability.

src/blr/cli/filterclusters.py
FrickTobias: I agree, however ```allMolecules``` and ```AllMolecules``` are an object (sometimes referred to as class) and an instance of that class which should be . Instance variables are those defined as ```self.<variable_name>```, which I believe all are named in lower case and separated by ```_```. 

However, I believe class methods are underlying functions to these classes, which currently has a similar naming system which should be changed (e.g ```addRead``` should be ```addread```).

I'm unable to find anything on naming conventions of the instances (the ```allMolecules```) so I thought it best to leave it for now.

src/blr/cli/filterclusters.py
FrickTobias: Doesn't sound unreasonable but where is it written? Just me being curious. 

Either way, I'll change it.

src/blr/cli/filterclusters.py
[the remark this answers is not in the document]
FrickTobias: Testing suggestion feature:

```suggestion
        with pysam.AlignmentFile(args.x2_bam, "rb") as openin, \
     pysam.AlignmentFile(args.output, "wb", template=openin) as openout:
```

src/blr/cli/filterclusters.py
FrickTobias:
```suggestion
    parser.add_argument("-M", "--max_molecules", metavar="<INTEGER>", type=int, default=500,
```

pontushojer: > Doesn't sound unreasonable but where is it written? Just me being curious.
> 
> Either way, I'll change it.

I am looking at this:
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

src/blr/cli/filterclusters.py
pontushojer: I was using the wrong wording with 'instance varliables' but I think that I am right in that `allMolecules` is a variable that stores an instance of the `AllMolecules`. As such it is variable and should follow those conventions. Am I making sense??

FrickTobias: Yeah that sounds about right. I also found one reference to mixedCapitalization which is supposed to be avoided at all costs, except for backwards compatibility so it's probably best to avoid it.

src/blr/cli/filterclusters.py
FrickTobias: Ahhhh, yes of course >.<

FrickTobias:
I implemented the suggested changes.

Issue #21, opened 2019-06-03
Performance improvements


marcelm:
I am collecting here some ideas for how to improve the performance of the pipeline.

* [X] On UPPMAX, `/scratch/` should be used for temporary files. It is faster than writing to the network file system. (See #115)
* [ ] More streaming should be employed. Currently, many huge intermediate FASTQ files are generated, some uncompressed, some compressed. This is slow: When writing them uncompressed, file I/O performance is likely a bottleneck. When writing them compressed, that takes a lot of CPU hours. Instead, one can let one program write uncompressed and send the output to the next program via a pipe. This avoids the intermediate file to disk entirely.
* [x] The `bc_extract.py` script should be rewritten so that it can work in a streaming fashion.
* [x] It should be possible to combine the last two Cutadapt commands into one.
* [x] Check whether it is faster to run multiple CD-HITs in parallel (instead of giving each CD-HIT the full number of cores).
* [x] Check whether a different clustering method (perhaps a custom one) is faster than CD-HIT (see #109)
* [x] Avoid the intermediate `samtools view` step (SAM can be piped directly into `samtools sort`).
* [x] Use FIFO tricks to run `samtools index` at the same time as generating the BAM that is being indexed. Again, this avoids I/O.
* [x] Test other read mappers (BWA-MEM, minimap2) (See #111 and #112)
* [x] Test alternatives to MarkDuplicates (`sambamba markdup`, `samblaster`) (Moved to #113)

marcelm:
A full pipeline run on a 29+40GB dataset (using 20 cores):

Stage | Wall-clock time (minutes) | Percentage
--- | ---: | ---:
Cutadapt 1  | 40 | 0.98%|
Barcode extraction  | 120 | 2.93%
Cutadapt 2  | 70 | 1.71%
Cutadapt 3  | 65 | 1.59%
Clustering preparation  | 45 | 1.10%
Clustering  | 900 | 21.98%
Read mapping  | 600 | 14.65%
BAM sorting  | 255 | 6.23%
BAM tagging  | 260 | 6.35%
Duplicate removal  | 300 | 7.33%
Duplicate marking  | 170 | 4.15%
Cluster merging  | 780 | 19.05%
BAM indexing  | 10 | 0.24%
Cluster filtering  | 240 | 5.86%
FASTQ generation  | 240 | 5.86%
**Total** | **4095** | 100.00%

FrickTobias:
[...]

PR #53, opened 2019-07-01
Fixed extractbarcodes so it works with fastq streaming

pontushojer:
*Related to #21*

Changed how the input/output is handle to allow interleaved files that can be used for streaming with cutadapt. Currently interleaved files are not used but I have done some test and it works with streaming using the snakemake pipe-flag.

**Changes:**
- Used `dnaio` instead of the FileReader class in `utils.py` for reading files and also for writing files.  
- New functions for checking input/output files.

### DIFF
--- a/src/blr/cli/extractbarcode.py
+++ b/src/blr/cli/extractbarcode.py
@@ -4,10 +4,59 @@ separated by an underline.
 """
 import logging
 import sys
+import dnaio
 
 import blr.utils as BLR
 
 logger = logging.getLogger(__name__)
+
+
+def input_files(inputs, interleaved):
+    """
+    Checks input arguments to see if they are sound and assigns them to input1 and input2
+    :param inputs: list of strings. Input files or paths given.
+    :param interleaved: boolean. If input is interleaved or not.
+    :return:
+    """
+    input1, input2 = (None, None)
+    if len(inputs) == 0:
+        sys.exit(f'EXITING: No inputs given!')
+    elif len(inputs) > 2:
+        sys.exit(f'EXITING: Too many files given as input {" ".join(inputs)}')
+    elif len(inputs) == 2:
+        input1, input2 = inputs
+        if interleaved:
+            interleaved = False
+        logger.info(f'Extracting barocodes from fastq {input1} and {input2}')
+    elif len(inputs) == 1 and interleaved:
+        input1 = inputs[0]
+        logger.info(f'Extracting barocodes from interleaved fastq {input1}')
+    else:
+        sys.exit(f'EXITING: Files given as input ({inputs}) does not match the required 1 or 2.')
+
+    return input1, input2, interleaved
+
+
+def output_files(output1, output2):
+    """
+    Checks output arguments given and if file should be written as interleaved or not.
+    :param output1: string. Output1 argument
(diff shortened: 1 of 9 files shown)


src/blr/cli/extractbarcode.py
```diff
@@ -3,45 +3,110 @@
...
 
 
 def add_arguments(parser):
-    parser.add_argument("r1", help="Read 1 fastq file")
-    parser.add_argument("r2", help="Read 2 fastq file")
-    parser.add_argument("out_r1", help="Read 1 output")
-    parser.add_argument("out_r2", help="Read 2 output")
+    # Files
+    parser.add_argument("input", nargs='*',
+                        help="Input paired fastq for read1 and read2. Use '-' if using stdin.")
```
FrickTobias: Since the function at least requires one argument I would write it as 
```
parser.add_argument("input1", help="Input paired fastq for read1 or interleaved file. Use '-' if using stdin.")
parser.add_argument("input2", nargs='?', help="Input paired fastq for read2. Use '-' if using stdin.")
```
This will replace all error handling in the ```input_files``` function and give standardized error messages from argparse for faulty user input.

FrickTobias: Subsequently one could remove the ```input_file``` function with 
```
if not args.input2: 
    interleaved = input1
    input1 = None
```
in the main function instead. One could also write ```if not args.input2 and args.interleaved:``` for error handling if we believe users might try and input single-end sequences.

src/blr/cli/extractbarcode.py
```diff
@@ -3,45 +3,110 @@
...
+    if output1 and output2:
+        interleaved = False
+        logger.info(f'Writing paired fastq to files {output1} and {output2}')
+    elif not output1 and not output2:
+        output1 = sys.stdout.buffer
+        interleaved = True
+        logger.info(f'Writing interleaved fastq to stdout.')
+    elif output1 and not output2:
+        interleaved = True
+        logger.info(f'Writing interleaved fastq to file {output1}.')
```
FrickTobias: I believe this can be  simplified by writing
```suggestion
if output2:
    interleaved = False
    logger.info(f'Writing paired fastq to files {output1} and {output2}')
else:
    interleaved = True 
    if output1:
        output1 = sys.stdout.buffer
        logger.info(f'Writing interleaved fastq to stdout.')
    else:
        logger.info(f'Writing interleaved fastq to file {output1}.')
```

FrickTobias: Hm, can't seem to get the multiline suggestion feature to work, either way this would replace lines 48-57.

pontushojer:
> Seems nice!
> 
> I have some suggestions on some improvements and some
> 
> * So far every script has the main script at the top so I think it's good to be consistent
> * I think it's good if we can give error messages regarding arguments through argparse, that way it'll look the same for all scripts (see suggestions).
> * Some if-statement checks were being done multiple times for the same loop which we can simplify (see suggestions)
> 
> I'm also thinking about it we should just assume the input is interleaved if only one input is given, since we need the input to be paired. This also makes sense with assuming interleaved if only one output file is specified.

Made new commit with these things in mind.

PR #57, opened 2019-07-04
Refactor tagbam

pontushojer:
- Cleaned up code. 
- Implemented tqdm for progress report.

@FrickTobias: One question, the option `-s, --skip_nonclust`, is that still relevant?

### DIFF
--- a/BLR_automation.sh
+++ b/BLR_automation.sh
@@ -42,6 +42,10 @@ threshold=0
 start_step=1
 end_step=4
 
+# BAM tags used.
+cluster_tag="BC"    # Used to store barcode cluster id in bam file. 'BX' is 10x genomic default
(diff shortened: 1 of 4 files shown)


pontushojer:
> No, `--skip_nonclust` is completely redundant at this point so feel free to remove it. It also follows that `seqs_in_cluster` variables and all related checks can be removed.
> 
> I would also suggest we add the option to use other tags (with default set to BC) to not have a hardcoded `set_tag('BC'<...>)`. There should be examples of this in e.g. `filterclusters` so we can keep consistent help messages.

Updated in new commit.

FrickTobias:
Great. Optimally I think the change of standard tag would have been a separate branch but I'll go ahead and name it accordingly in the merge instead for this time since it also just a convention related change rather than a functional one.

Issue #59, opened 2019-08-01
filtercluster stats for output molecules_per_bc is wrong


pontushojer:
I just noticed that the statistics file outputed by the flitercluster scripts is currently wrong. Running on the test files produces the following output:

```
(blr) PontusHMBA:BLR pontus.hojer$ cat outdir/cluster_stats/x2.stats.molecules_per_bc
11112212121222111112221221211112211111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
```

I could not see any immediate errors in the script so maybe we have to look deeper into it if its a file we would like to keep.

PR #60, opened 2019-08-05
bugfix: no newline in mol/bc out

FrickTobias:
Resolving issue #59

### DIFF
--- a/src/blr/cli/filterclusters.py
+++ b/src/blr/cli/filterclusters.py
@@ -279,7 +279,7 @@ class Summary:
             number_of_molecules = len(molecule_dict[barcode])
             print(f"{number_of_molecules}", file=molecules_per_bc)
             for molecule in (molecule_dict[barcode]):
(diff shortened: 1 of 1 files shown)


src/blr/cli/filterclusters.py
```diff
@@ -277,7 +277,7 @@ def write_molecule_stats(self, output_prefix, molecule_dict):
         # Writing molecule-dependant stats
         for barcode in tqdm(molecule_dict):
             number_of_molecules = len(molecule_dict[barcode])
-            molecules_per_bc.write(str(number_of_molecules))
+            molecules_per_bc.write(str(number_of_molecules) + "\n")
```
marcelm: This would be a good place to use `print` instead. Like this:
```
print(number_of_molecules, file=molecules_per_bc)
```
That way, you avoid the explicit `\n` and also `str()`.

FrickTobias: Cool, never used print for writing to files before.

PR #61, opened 2019-08-05
Removed intermediate samtools view step

pontushojer:
Related to #21 checkpoint: "Avoid the intermediate samtools view step (SAM can be piped directly into samtools sort)."

Removed intermediate samtools view step and instead merge with next-coming sorting step.


FrickTobias:
Looks good to me!

PR #62, opened 2019-08-07
Trimming files in pipe instead of sequentially

pontushojer:
Related to #21 . 

By removing sequential trimming steps and compression we can increase performance. This was done using snakemakes pipe() function. Note that this required a minimum of 3 cores for the script to be able to run.

I also moved the trimming rules from the Snakefile to  `rules/trim.smk` to increase modularization of the Snakefile.


rules/trim.smk
```diff
@@ -0,0 +1,67 @@
...
+
+rule trim_r1_handle_pipe:
+    #Trim away E handle on R1 5'. Also removes reads shorter than 85 bp.
+    output:
+        interleaved_fastq=pipe("{dir}/trimmed-a.fastq")
+    input:
+        r1_fastq="{dir}/reads.1.fastq.gz",
+        r2_fastq="{dir}/reads.2.fastq.gz"
+    log: "{dir}/trimmed-a.log"
+    threads: int(available_cpu_count()/3)
```
FrickTobias: Won't this mean that if any non-multuple of 3 is supplied as number of cores leaves us with non-used cores (since `int()` truncates the values (e.g. supplying 5 cores will give only use 3). Is there a way to use all of them?
 
I'm thinking something like [this](https://stackoverflow.com/questions/20348717/algo-for-dividing-a-number-into-almost-equal-whole-numbers/20348992):
```
num, div = 15, 4
print ([num // div + (1 if x < num % div else 0)  for x in range (div)])
```

pontushojer: Interesting! I just adopted this from a issue i found on bitbucket but there is for sure a more efficient way to utilize the available cores. 

As extractbarcodes only runs on a single core I adjusted it as follows:

Step | Threads | 5 cores total | 6 cores total
--- | --- | ---: | ---:
trim_a | `math.ceil((available_cpu_count()-1)/2)` | 2 | 3
extractbarcodes | 1 | 1 | 1
trim_c | `int((available_cpu_count()-1)/2)` | 2 | 2

pontushojer:
@FrickTobias what about this?

rules/trim.smk
FrickTobias: That's true, extractbarcodes only uses one core and we can revisit this if we switch to cutadapt for barcode recognition later (since that'll be able to thread the process).

PR #63, opened 2019-08-09
Block quotes

pontushojer:
Moved from line quotes to block quotes where appropriate. Some small cleanups and other style changes included.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -30,10 +30,12 @@ if config["index_nucleotides"] == 0:
             stdout = "{dir}/cdhit_prep.stdout",
             stderr = "{dir}/cdhit_prep.stderr"
         shell:
-            "blr cdhitprep "
-            " {input.r1_fastq}"
-            " {output}"
-            " -f 0 > {log.stdout} 2> {log.stderr}"
+            """
+            blr cdhitprep \
+             {input.r1_fastq} \
+             {output} \
+             -f 0 > {log.stdout} 2> {log.stderr}
+            """
 else:
     rule cdhitprep:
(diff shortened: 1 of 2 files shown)


marcelm:
I would prefer for this not to be changed. The reason I used normal quotes instead of triple quotes is that they make the output from `snakemake -p` much nicer because they don’t introduce stray spaces.

With this Snakefile:
```
rule:
    shell:
        """
        echo \
        hello
        """
```
you get this output from `snakemake -p`:
```
rule 1:
    jobid: 0


        echo         hello
        
hello
```
Using double quotes instead:
```
rule:
    shell:
        "echo"
        " hello"
```
you get this output:
```
rule 1:
    jobid: 0

echo hello
hello
```
This gets worse the more line continuations you have.

I almost always use `-p` and was planning on making it the default in BLR. It’s very useful to be able to see what is going on. For example, when debugging, I often need to copy and paste commands from the terminal and run them manually. Extraneous spaces are a nuisance.

With block quotes, you do save two quote characters on each line, but since you add one backslash, the advantage isn’t so big.

Personally, I also find the style nicer: All lines that are continuations of the same command are indented by one space such that you can see that the lines belong to the command. Since the spaces all line, I find this easier to read than the backslash-style, where the backslashes are at the end of the line and usually not aligned.

I agree that triple quotes are more appropriate when multiple commands are run in a single rule because then you need actual line breaks.

pontushojer:
I have not used the -p option much but it does seem useful. It was not something we considered when we talked about making this change. As you talked about it was mostly a style base change. In that case I agree that we drop this change. I will however include some of the changes not relating to block-quotes in a new PR.

PR #64, opened 2019-08-12
Style changes from closed PR block-quote.

pontushojer:
Changes form #63 not related to block quotes. 

Includes:
- Changed naming for some things
- Removed parenthesis for log directions
- Made style of shell commands more similar between rules.


FrickTobias:
Looks good.

PR #66, opened 2019-08-15
Remove old automation scripts

pontushojer:
These are not currently used and not even compatible with the current snakemake pipeline. There is however some things not yet included in the snakefile such as this marked TODO:

```
    # Ugly solution to calculate % construOK	
    # TODO	
#    var1=$( cat $trim_logfile | grep 'Read 1 with adapter' | cut -d '(' -f 2 | cut -d '%' -f 1 | tr '\n' ' ' | cut -d ' ' -f 1 )	
#    var2=$( cat $trim_logfile | grep 'Read 1 with adapter' | cut -d '(' -f 2 | cut -d '%' -f 1 | tr '\n' ' ' | cut -d ' ' -f 2 )	
#    printf "`date`""\t"; awk '{print "Intact reads: "$1*$2*0.0001" %"}' <<< "$var1 $var2"
``` 

Is this something we want to keep ?


FrickTobias:
I think it is fine, we have those stats in the log files.

Issue #67, opened 2019-08-15
Maximum available threads requested from piped rules. 


pontushojer:
So I have noticed a recurring error which I think relates to the piped rules I introduced in out rules/trim.smk file. Since there are three rules connected through the pipe, snakemake requires a minimum of three threads to run the trimming. Problem is that snakemake does not automatically adjust this for the three rules so that if i set `threads: 20` for each of the three rules and give three threads the first rule will take all three and snakemake raise a WorkflowError requesting five threads (one more for each other rule). This is already noticed in a current [snakemake issue on bitbucket]( https://bitbucket.org/snakemake/snakemake/issues/1039/split-number-of-threads-for-a-rule).

In my solution I used the same solution as in the issue to cap the number of threads requested by the rule using the `available_cpu_count` function from [snakemake utils](https://snakemake.readthedocs.io/en/stable/api_reference/snakemake_utils.html#snakemake.utils.available_cpu_count) with some modifications. Currently it looks as follows: 

```yaml
rule trim_r1_handle_pipe:
    ...
    threads: ceil((available_cpu_count()-1)/2)
    ...

rule extract_barcodes_pipe:
    ...
    (Only runs one thread)
    ...

rule final_trim_pipe:
    ...
    threads: int((available_cpu_count()-1)/2)
    ...
```

I thought this function would give the actual number of threads given in the argument (`snakemake -j`) but instead this is based on the total number of available CPUs on the local machine. Thus it requires that you run the pipeline with the full number of CPUs or it will raise an error such as this (from our local server with 24 CPUs:

```
WorkflowError:
Job needs threads=24 but only threads=20 are available. This is likely because two jobs are connected via a pipe and have to run simultaneously. Consider providing more resourc
es (e.g. via --cores).
```

I have tested reading the arguments given in the command i.e. taking whatever comes after `-j`in the sys.argv list, but this seems as a bad solution. Alternatively I could move from using the snakemake pipe function on three rules to a single rule which uses regular unix pipes (i.e. `|`).

FrickTobias:
Why is it bad to use the `threads` variable from the `-j` option? 

Anyways, have you tested looking into regulating the number of cores given to each individual pipe command, ie instead of writing
```
rule trim_r1_handle_pipe:
    ...
    threads: ceil((available_cpu_count()-1)/2)
    shell:
        "cutadapt"
        ...
        " -j {threads}"
        ...
```
we could write something along the lines of
```
rule trim_r1_handle_pipe:
[...]

PR #69, opened 2019-08-19
Snakemake piped rules merged into single rule 

pontushojer:
Fix to issue #67 

Joined three piped rules into one and used standard pipes. This is a bit harder to read as the shell command is so massive. However it does not have the same problems with thread distribution as using the snakemake pipe() functionality.


FrickTobias:
I agree that this will have to suffice for now.

Issue #18, opened 2019-06-03
SamToFastq fails because of unpaired mates


marcelm:
When running the pipeline (on the test data), during the very last step when `SamToFastq` is used to convert the final BAM to FASTQ, the following error occurs:
```
Exception in thread "main" htsjdk.samtools.SAMFormatException: SAM validation error: ERROR: Found 493 unpaired mates
```
I found a forum post by the author of Picard somewhere (sorry, lost the link) stating that this validation step occurs *after* the output FASTQ has already been written, which means that the error doesn’t become apparent unless one looks at the log file or checks the exit code.

The problem is that MarkDuplicates is run with `REMOVE_DUPLICATES=true`, which apparently deletes single reads out of pairs.

This can be “fixed” by adding `VALIDATION_STRINGENCY=silent` to SamToFastq, but that’s just hiding the problem. Usually, MarkDuplicates is indeed used to only mark duplicates, not to really remove them, in order to avoid losing data. The idea is that one can then (in theory) keep only a single BAM file and be able to reconstruct the full input FASTQ file from it. My preferred course of action would be to follow this best practice, but I don’t know how the created FASTQ and BAM files are actually used later.

FrickTobias:
In an optimal solution, duplicates would not be removed here because of the problem you correctly point out with single reads out of pairs.

However, currently read duplicates are used for finding barcode duplicates (aka where two barcode has tagged the same molecule). This leads to two barcodes sharing a lot of proximal read duplicates which subsequently are merged into one barcode for all reads sharing that barcode sequence. 

The actual procedure doing this is first to remove all duplicates with the BARCODE_TAG=BC option and  afterwards analyze duplicate reads left in the file with cluster_rmdup.py (by first marking duplicates without the BARCODE_TAG flag).

I suggest that I rewrite the script to so it ignores duplicate reads at the same positions if they do not have different barcodes, essentially compacting the duplicate removal (BARCODE_TAG=BC), duplicate marking & cluster_rmdup step into duplicate removal & cluster_rmdup.

PR #72, opened 2019-08-27
Keep read duplicates

FrickTobias:
Solving #18. Only mark duplicates instead of removing and then marking to find barcode duplicates. This included:

- Removal of read duplicate removal set
- clusterrmdup now requires at least two barcodes at a duplicate position to save it in cache for later analysis (not needed before since it could be assumed)
- filterrclusters 
- Changed md5sum to match
- Removed silence option from SamToFastq step

Since md5sum was not comparable before and after, the last bam files coverage graph (samtools depth output) md5sum was compared instead.

All changes except the last were mistakenly included into PR #68 and instead I'll write the names here. Commit names not included in this PR:

Aug 26, 2019
- Change md5sum (duplicates left in file & other minor changes)

Aug 20, 2019
- Adjust name to fit snakefile (keep read dup+namechange x2 => bcmerge)
- Add row to skip reads marked as duplicates in file
- Require at least two different barcodes at a given rp position
- Remove picard read rmdup step and adjust names (also changed from x2 …

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -193,5 +193,4 @@ rule bam_to_fastq:
         "{params.picard_command} -Xms{params.heap_space}g SamToFastq"
         " I={input.bam}"
         " FASTQ={output.r1_fastq}"
-        " VALIDATION_STRINGENCY=SILENT"
         " SECOND_END_FASTQ={output.r2_fastq} 2>> {log}"


FrickTobias:
Instead of removing read duplicates and only marking those with different barcodes we now only mark them (regardless of barcode sequence). This does not only keep the duplicates but also affects the `clusterrmdup` and the `filterclusters` step, which is why I compared the md5sum of the coverage graph (which turned out to be unchanged) to ensure the change didn't do anything unintentional.

PR #73, opened 2019-08-28
Write settings

FrickTobias:
In order to always know what settings have been used I added some rows to main to it always starts with writing all things defined from args, I also added the levelname to logger so one can see which submodule is producing the messages.

With an increasing amount of options this might become a quite massive string, but we don't have to many options yet and in that case I'd throw in a check so it only prints the non-default arguments used (but havn't quite figured out how to do that yet).

I also fixed the log going to 4_rmdup.log which were all overwriting each other.

Example of output:
```
main - SETTINGS: input_tagged_bam: mapped.sorted.tag.mkdup.bam
main - SETTINGS: output_bam: mapped.sorted.tag.mkdup.bcmerge.bam
main - SETTINGS: merge_log: barcode-merges.csv
main - SETTINGS: barcode_tag: BC
main - SETTINGS: window: 100000
main - SETTINGS: module: <module 'blr.cli.clusterrmdup' from '/Users/tobiasfrick/PycharmProjects/masters/BLR/src/blr/cli/clusterrmdup.py'>
clusterrmdup - INFO: Starting Analysis
Processing reads: 19976it [00:00, 38105.78it/s]
Writing output: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 19976/19976 [00:00<00:00, 118471.67it/s]
clusterrmdup - INFO: Total Reads in file: 19976
clusterrmdup - INFO:   Unmapped reads: 1270
clusterrmdup - INFO:   Reads with unmapped mate: 1102
clusterrmdup - INFO:   Reads without BC tag: 0
clusterrmdup - INFO:   Reads analyzed: 17604
clusterrmdup - INFO: Reads at dup pos: 3468
clusterrmdup - INFO: Reads with new tag: 3696
clusterrmdup - INFO: BC tags removed: 268
clusterrmdup - INFO: Finished
```

### DIFF
--- a/src/blr/__main__.py
+++ b/src/blr/__main__.py
@@ -34,6 +34,8 @@ def main() -> int:
     if not hasattr(args, "module"):
         parser.error("Please provide the name of a subcommand to run")
     else:
+        for object_variable, value in vars(args).items():
+            logger.info(f"{object_variable}: {value}")
         module = args.module
         del args.module
         module.main(args)
--- a/src/blr/__main__.py
+++ b/src/blr/__main__.py
@@ -35,7 +35,7 @@ def main() -> int:
         parser.error("Please provide the name of a subcommand to run")
     else:
         for object_variable, value in vars(args).items():
-            logger.info(f"{object_variable}: {value}")
+            sys.stderr.write(f"main - SETTINGS: {object_variable}: {value}\n")
         module = args.module
         del args.module
         module.main(args)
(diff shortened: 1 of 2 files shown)


src/blr/__main__.py
```diff
@@ -34,6 +34,8 @@ def main() -> int:
     if not hasattr(args, "module"):
         parser.error("Please provide the name of a subcommand to run")
     else:
+        for object_variable, value in vars(args).items():
+            sys.stderr.write(f"main - SETTINGS: {object_variable}: {value}\n")
```
pontushojer:
```suggestion
            sys.stderr.write(f" {object_variable}: {value}\n")
```

Issue #74, opened 2019-09-03
Include frameshift constructs


pontushojer:
We have started testing frameshift constructs for the BLR method. This means that the first handle might be preceded by 0, 1, 2 or 3 nucleotides. We should investigate how this affects the trimming and if we have to edit the trimming somehow to allow for this new setup.

FrickTobias:
By just removing the anchoring we should reach similar results.

marcelm:
Because just leaving out the anchoring gives you false positive matches in the middle of the read, I’ve added so-called [non-internal 5' adapters](https://cutadapt.readthedocs.io/en/stable/guide.html#non-internal-5-and-3-adapters) to Cutadapt a while ago. Instead of `-g ^HANDLE`, you would write `-g XNNNHANDLE` in this case. The `X` prevents the match from occurring in the middle of the read, but allows partial matches in the beginning. (And because of the three `N`s, up to three preceding nucleotides.) If you also set the minimum overlap length to the length of the handle, this will work as required.

pontushojer:
This issue was resolved in PR #82.

PR #75, opened 2019-09-04
Allow frameshift adaptors in constructs

pontushojer:
Fix for issue #74.


FrickTobias:
I'm not sure we should add the length requirement. My thought is that requiring ~65 (25+40) bases of construct with a max 20 % error rate should suffice in terms of stringency.

Otherwise it looks great though :)

marcelm:
If you don’t add `min_overlap=26`, then you can get partial matches of the 3' end of the adapter to the 5' end of the read. Because the default is to require an overlap of three bases, simply finding `TGG` in the beginning of a read would be enough for a match.

FrickTobias:
Okay, so that is essentially the minimum of matches needed in order to proceed to the alignment-error calculation or am I mistaken? If so, what are the initial requirement of the alignemnt?

marcelm:
The minimum overlap is the minimum number of bases of the adapter that need to be covered by the alignment. These don’t have to be matches necessarily. I don’t know what you mean by initial requirements.

FrickTobias:
Alright. So essentially `min_overlap=26` with `-e 0.2` means we can have an alignment of 26 bases of 26 bases where 5 bases can be inserted or substituted but none can be deleted? 

My thought a deletion would cause the alignment length to be less than 26? Or possibly mistakenly trimming a base from the the next part of the read constructs to increase the length to 26 (if it can be tolerated within the 20% error rate).

I don't know if I'm missing something but currently requiring a minimum length equal to the total length of the construct seem quite stringent. 

PS. Never mind my question about initial requirement, didn't think properly.

marcelm:
> So essentially min_overlap=26 with -e 0.2 means we can have an alignment of 26 bases of 26 bases where 5 bases can be inserted or substituted but none can be deleted?

You can have in total 5 errors, and any of them can be substitutions, insertions or deletions.

The overlap length is computed as the number of adapter bases involved in the alignment. It is not the length of the alignment, precisely because of the issue you mentioned: I don’t want the length to change just because there is an insertion or deletion.

So assume you have the 5' adapter `ANICEADAPTER`. Here is the alignment to a read with a deletion:
```
ANICEADAPTER (adapter)
     ==X====
     AD-PTERANDTHENTHEREST (read)
```
And to a read with an insertion:
```
ANICEADA-PTER (adapter)
     ===X===
     ADARPTERANDTHENTHEREST (read)
```
In both cases, the overlap length is 7. If the error rate allows 1/7 errors (ca 15%), then both alignments will be found.

> I don't know if I'm missing something but currently requiring a minimum length equal to the total length of the construct seem quite stringent.

Use the overlap length to describe what you would expect to see in the read if there were no sequencing errors. If the adapter sequence could be degraded at its 5' end due to a step in your library preparation, then you should use a lower value than the actual adapter length to allow this. If you expect the full sequence to be present, then use `min_overlap=26`. The error rate will then take care of any sequencing errors.

FrickTobias:
Okay, sounds great! Thanks for the explanation!

PR #76, opened 2019-09-05
Tag restructure

FrickTobias:
I've made the molecule building part of `filterclusters` a separate step I call `buildmolecules` which instead of only building a dictionary in memory tags the bam file with molecule ID:s (MI, coherent with the 10x convetions) as well as a tag (MN) for barcodes which stores the number of molecules in the for that specific barcode.

This will then make it possible to reach molecule information without recreating it, mainly thought to  be needed for when we will be phasing linked reads based on phased reads from informative positions (hetSNP sites).

Since we now have more tags relating barcoding information `filterclusters` now takes a list of tags which to be removed from the read. Currently only the barcode flag is removed in the pipe, but I thought I could fix this by adding it to our config file setup while the rest is being reviewed.


FrickTobias:
Yes, the script is not filtering reads and this is because they are still intact genomic information, we just can't trust the phasing information stored in them. Consider a bubble with 500 molecules with 50 reads each which all have been tagged with the same droplet barcode. We cannot trust (given a 30kb molecule length) that their barcodes do not collide so we cannot assume proximal reads are origin to the same molecule, however they can still be considered normal 25,000 non-linked shotgun sequencing reads (which can be used for e.g. SNV variant calling). Regardless, I would still only mark them just as we're now keeping read duplicates (but marking them as such).

The actual molecule length is unknown prior to building them and this average is dependent on the --window argument. We investigated this and concluded 30 kb (molecule length was roughly 25 kb) was a quite robust window size, but I agree having some kind of dynamic read depths would probably be an improvement on this. 

I think moving those to utils sound like great ideas! But as you point out I'm thinking a separate PR is probably appropriate.

PR #80, opened 2019-09-09
Phasing pipeline 

pontushojer:
This is the initial work on the phasing pipeline. I have used [HapCUT2](https://github.com/vibansal/HapCUT2) for phasing according to the [10x Genomics linked-reads protocol](https://github.com/vibansal/HapCUT2#10x-genomics-linked-reads). Hapcut2 is currently being installed through a bash script `install-hapcut2.sh` as no conda release is available. 

For reference VCF file I have included two option so far. Either you can call variants using [freebayes](https://github.com/ekg/freebayes) or used a high confidence set. For the tests runs I have been using a high confidence set from [GIAB for HG002_NA24385 genome](https://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/release/AshkenazimTrio/HG002_NA24385_son/NISTv3.3.2/GRCh38/supplementaryFiles/) which we have used for the testdata. This means that I have also updated the current testdata to `blr-testdata-0.2`. 

Other changes included:
- Changed from BC to BX for barcode tag (this is required for HapCUT2)
- Rule for checking phased vcf against "ground truth" phased vcf which is also provided as test data. 
 - Output from filterclusters  is now indexed as this is needed for LinkFragments script in HapCUT2 phasing.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -233,7 +233,7 @@ rule HAPCUT2_extractHAIRS:
         vcf = "{dir}/reference.vcf"
     log: "{dir}/hapcut2_extracthairs.log"
     shell:
-         "{HAPCUT2}/build/extractHAIRS"
+         "config[hapcut2]/build/extractHAIRS"
          " --10X 1"
          " --bam {input.bam}"
          " --VCF {input.vcf}"
@@ -248,7 +248,7 @@ rule HAPCUT2_LinkFragments:
         unlinked = "{dir}/mapped.sorted.tag.rmdup.x2.filt.unlinked"
     log: "{dir}/hapcut2_linkfragments.log"
     shell:
-         "python {HAPCUT2}/utilities/LinkFragments.py"
+         "python config[hapcut2]/utilities/LinkFragments.py"
          " --bam {input.bam}"
          " -v {input.vcf}"
          " --fragments {input.unlinked}"
@@ -263,7 +263,7 @@ rule HAPCUT2_phasing:
         vcf = "{dir}/reference.vcf"
     log: "{dir}/hapcut2_phasing.log"
     shell:
-         "{HAPCUT2}/build/HAPCUT2"
+         "{config[hapcut2]}/build/HAPCUT2"
          " --nf 1"
          " --fragments {input.linked}"
          " --vcf {input.vcf}"
--- a/Snakefile
+++ b/Snakefile
@@ -181,16 +181,21 @@ rule filterclusters:
     # Filter clusters based on parameters
     output:
         bam = "{dir}/mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam",
+        bai = "{dir}/mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam.bai"
     input:
         bam = "{dir}/mapped.sorted.tag.mkdup.bcmerge.mol.bam"
     log: "{dir}/filterclusters.log"
     shell:
         "blr filterclusters"
         " {input.bam}"
-        " {output.bam}"
+        " -"
         " -mn {config[num_mol_tag]}"
         " -M 260"
-        " -t {config[cluster_tag]} {config[molecule_tag]} {config[num_mol_tag]} {config[sequence_tag]} 2> {log}"
+        " -t {config[cluster_tag]} {config[molecule_tag]} {config[num_mol_tag]} {config[sequence_tag]}"
+        " 2>> {log} |"
+        " tee {output.bam} |"
+        " samtools index  - {output.bai}"
+
 
 rule bam_to_fastq:
     # Convert final bam file to fastq files for read 1 and 2
--- a/Snakefile
+++ b/Snakefile
@@ -1,4 +1,4 @@
-from snakemake.utils import validate
+from snakemake.utils import validate, WorkflowError
 import itertools
 import os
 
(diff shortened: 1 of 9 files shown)


Snakefile
```diff
@@ -203,3 +211,27 @@ rule bam_to_fastq:
...
+        input:
+             bam = "{dir}/mapped.sorted.tag.mkdup.bcmerge.filt.bam"
+        log: "{dir}/call_variants_freebayes.log"
+        params:
+            reference = config["bowtie2_reference"] + ".fasta" # I am unsure if this is a good solution, but it works.
+        shell:
+             "freebayes"
+             " -f {params.reference}"
+             " {input.bam} 1> {output.vcf} 2> {log}"
+elif os.path.exists(config['reference_variants']):
```
FrickTobias: If a path is supplied but is given erroneously the `elif` statement the error message would be confounding since it also checks if path exists. I think a better solution is to only check for the argument and letting snakemake point out any non-existing paths (as we would for e.g. missing read files).

```suggestion
elif config['reference_variants']:
```

Snakefile
```diff
@@ -203,3 +211,27 @@ rule bam_to_fastq:
         " I={input.bam}"
         " FASTQ={output.r1_fastq}"
         " SECOND_END_FASTQ={output.r2_fastq} 2>> {log}"
+
+if config['call_variants']:
```
FrickTobias: Personally I would default to calling variants since if no reference file is supplied I think it is safe to assume the user wants to call variants on the sample. To encourage users to supply a reference when available, a basic usage or best practices example can be provided in the README.

Then the structure would roughly follow:
```
if config[reference]:
    <do things>
else:
    <call_variants>
```

rules/phasing.smk
```diff
@@ -0,0 +1,59 @@
+HAPCUT2=config["hapcut2"]
+
+rule HAPCUT2_extractHAIRS:
```
FrickTobias: I'm not sure if I remember correctly but I believe rules were functions and therefore should be written with underline separation according to [pep8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).
```suggestion
rule hapcut2_extract_hairs:
```

rules/phasing.smk
```diff
@@ -0,0 +1,59 @@
...
+        vcf = "{dir}/reference.vcf"
+    log: "{dir}/hapcut2_extracthairs.log"
+    shell:
+         "{HAPCUT2}/build/extractHAIRS"
+         " --10X 1"
+         " --bam {input.bam}"
+         " --VCF {input.vcf}"
+         " --out {output.unlinked} 2> {log}"
+
+rule HAPCUT2_LinkFragments:
```
FrickTobias:
```suggestion
rule hapcut2_link_fragments:
```

rules/phasing.smk
```diff
@@ -0,0 +1,59 @@
...
+        unlinked = "{dir}/mapped.sorted.tag.mkdup.bcmerge.mol.filt.unlinked"
+    log: "{dir}/hapcut2_linkfragments.log"
+    shell:
+         "python {HAPCUT2}/utilities/LinkFragments.py"
+         " --bam {input.bam}"
+         " -v {input.vcf}"
+         " --fragments {input.unlinked}"
+         " --out {output.linked} &> {log}"
+
+rule HAPCUT2_phasing:
```
FrickTobias:
```suggestion
rule hapcut2_phasing:
```

rules/phasing.smk
```diff
@@ -0,0 +1,59 @@
...
+    log: "{dir}/hapcut2_phasing.log"
+    shell:
+         "{HAPCUT2}/build/HAPCUT2"
+         " --nf 1"
+         " --fragments {input.linked}"
+         " --vcf {input.vcf}"
+         " --out {output.phase}"
+         " --outvcf 1 2> {log}"
+
+rule HapCUT2_stats:
```
FrickTobias:
```suggestion
rule hapcut2_stats:
```

rules/phasing.smk
FrickTobias: And I see HapCUT2 doesn't follow these conventions (it's also not written in python) but I think we should follow pep8 to the best of out ability whenever possibility.

pontushojer: I am not sure I would agree that the naming of rules should **have to** follow python convention. These are not python objects as they are specific to snakemake. But I also don't care much either way. If you feel this would improve consistency we could set this as a practise for this project.

Snakefile
pontushojer: Yep, this is definitly simpler than the current logic, will change!

pontushojer:
> The only general comment/question I have is regarding the naming of the freebayes output files, do they have a specified format we could add to the end? If they are just plain text files maybe adding `.txt` would be appropriate or was this something taken from common practices from people working with freebayes?

You are talking about HapCUT2 here I persume?? And about the `.unlinked`, `.linked`and `.phase` files? In that case yes there is no harm adding the .txt extension, I was merely following the naming used in the HapCUT2 docs.

rules/phasing.smk
FrickTobias: I vaguely would like @marcelm mentioning something about snakemake rules being functions and since I can't find it now, he would have to remind me whether my memory serves me right or not.

Either way, looking at the [snakemake docs](https://snakemake.readthedocs.io/en/stable/) every example name is written as lower_case functions in python (assuming NAME is just written as a placeholder) so I'd prefer to follow that if there are no established conventions.

marcelm: They are functions under the hood, but if that weren’t the case, they kind of look like functions, so making them consistently lowercase would make sense.

Issue #85, opened 2019-09-12
Leaving out reference_variants configuration crashes


marcelm:
```
WorkflowError in line 6 of .../Snakefile:
Error validating config file.
ValidationError: None is not of type 'string'

Failed validating 'type' in schema['properties']['reference_variants']:
    OrderedDict([('type', 'string'),
                 ('description',
                  'Path to reference variants, if not provided then '
                  'variant will be called by freebayes'),
                 ('default', None)])

On instance['reference_variants']:
    None
```
The comment says that FreeBayes will be run if `reference_variants` is left out, but doing so leads to the above error.

PR #86, opened 2019-09-13
Fixed validation error 

pontushojer:
Fix for issue #85. Reference_variants defaults to None which fails validation (string required). Also updated this for other paths in config.schema.yaml and config.yaml.


config.schema.yaml
```diff
@@ -40,8 +40,8 @@ properties:
...
     type: string
     description: Path to reference variants, if not provided then variant will be called by freebayes
-    default:
+    default: ""
   phasing_ground_truth:
     type: string
     description: Path to phased variants used to compute phasing stats.
-    default: .
\ No newline at end of file
+    default: ""
```
marcelm: There’s a missing newline at the end of the file

pontushojer: Fixed in new commit.

marcelm:
I would suggest to fix the problem in another way: By changing the schema such that it allows `null` in addition to a `string` for the affected fields. According to https://stackoverflow.com/questions/16241333/ , you can use `type: "string, null"` to achieve this.

That way, one can just leave out the value (as your intention was from the beginning I gather) instead of having to specify an empty string with `""`.

pontushojer:
> I would suggest to fix the problem in another way: By changing the schema such that it allows `null` in addition to a `string` for the affected fields. According to https://stackoverflow.com/questions/16241333/ , you can use `"type": ["string", "null"]` or perhaps `"type": "string, null"` to achieve this.
> 
> That way, one can just leave out the value (as your intention was from the beginning I gather) instead of having to specify an empty string with `""`.
> 
> _Edit:_ Fixed suggested schema.

I did not know that was an option but that is indeed much nicer. Yes my intension was to just leave it out the value completely. 

Changed in new commit.

PR #95, opened 2019-09-16
Implement a "blr run" command

### LINKED ISSUE #92: Implement a "blr run" command or similar
It should not be required for the user to explicitly run `snakemake` (and to know the location of the Snakefile). Similar to the way IgDiscover works, we can have some kind of `blr init` command that initalizes an analysis directory with a configuration file that creates sane defaults (see #94). Then, calling `blr run` from within the directory would run the pipeline.

marcelm:
This is equivalent to running "snakemake -p", but without the need to specify the path to the Snakefile.  Closes #92

The first commit in this PR is actually #93, which should be merged first.


marcelm:
This is now ready to be merged (merge #93 first, though). There were lots of Travis failures because of a mismatching MD5 sum, but re-starting the build sufficiently often made it finally pass. We’ll need to investigate why the MD5 keeps changing, though.

PR #97, opened 2019-09-17
Use environment files for both abstract and locked dependencies

marcelm:
With this change, we will have two types of Conda environment files:

- The `environment.yml` file contains only abstract dependencies such as
  "picard=2.10" or "pysam".  This file is managed manually and needs to be
  updated whenever there are new dependencies or when the required version for
  dependencies change.

- The `environment.*.lock.yml` files contain a fully specified description of the
  entire environment (with locked-down versions).  This is needed because we
  want to have reproducible tests.  Use the script `misc/condalock.sh` to
  create these files whenever you make changes to `environment.yml`.


marcelm:
I believe the Travis failure is due to Cutadapt being upgraded from 2.3 to 2.5. Cutadapt 2.3 had a bug where it would ignore the `-e` (error rate) option for linked adapters, but because 2.5 fixes that, the results are sligthly different.

marcelm:
This is ready now: I’ve now pinned Cutadapt to 2.3 to circumvent the problem. We can update the version in a separate PR.

PR #102, opened 2019-09-23
Run Bowtie with --reorder

### LINKED ISSUE #101: Random MD5 sum mismatches
When running the tests, the MD5 sum of the `mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam` sometimes mismatches. This appears to be random and happens perhaps one out of 10 times.

marcelm:
This ensures that the order of records in Bowtie’s output is the same as the
order of reads in the input FASTQ files.  Without this, the order is more or
less random and we cannot rely on it in the tests.

Closes #101

### DIFF
--- a/src/blr/Snakefile
+++ b/src/blr/Snakefile
@@ -101,6 +101,7 @@ rule bowtie2_mapping:
     log: "bowtie2_mapping.log"
     shell:
         "bowtie2"
+        " --reorder"
         " -1 {input.r1_fastq}"
         " -2 {input.r2_fastq}"
         " -x {params.reference}"
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -19,8 +19,8 @@ cp tests/test_config.yaml outdir/config.yaml
 cd outdir
 blr run reads.1.final.fastq.gz reads.2.final.fastq.gz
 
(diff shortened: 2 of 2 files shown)


marcelm:
Since `--reorder` comes with a performance penalty according to the Bowtie documentation, I considered these alternatives:

- Before checking the MD5 sum, sort by both name and position. This should normalize the order to be the same regardless of what the input order was.

- Use `--threads=1` during the tests. When running Bowtie with only one thread, it will always produce correctly ordered output.

However, I find it a desirable property of a pipeline to always produce the same results, so in my opinion `--reorder` should also be enabled in a production version. On the other hand, we may later want to revisit this and benchmark what the slowdown actually is.

FrickTobias:
(That rule would then probably be easier to include and make mapper independent by doing something like your first suggestion where files are sorted just before md5sum checks)

FrickTobias:
Either way, we can implement such an option later if we see the run time it significantly increased (or if we change mapper).

marcelm:
I have now benchmarked `--reorder` vs not using `--reorder` and there is no difference.

Times for mapping 20 million reads, repeated three times:

* Without `--reorder`: 43m 33s, 43m 22s, 43m 27s
* With `--reorder`: 43m 20s, 43m 14s, 43m 13s

PR #103, opened 2019-09-24
Run flake8 as part of CI tests

marcelm:
This will run flake8 (a Python style checker) as part of the Continuous Integration tests. This tool finds things such as unused imports and variables or too long lines. This PR also fixes all the things it complains about. In the flake8 configuration, I’ve set the maximum line length to 119 in order to avoid having to reformat too much. This is a bit high in my opinion, and I’d prefer something around 100 or 90, but we can discuss this at a later time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,10 @@
-language: python
+language: generic
 
 dist: xenial
 
 cache:
   directories:
     - $HOME/.cache/pip
-
-python:
-  - "3.6"
-os:
-  - linux
-  - osx
 
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh; fi
@@ -37,7 +31,13 @@ script:
 
 jobs:
   include:
+    - name: Linux
+      os: linux
+    - name: macOS
+      os: osx
     - name: flake8
+      language: python
+      python: "3.6"
       before_install:
       install: python3 -m pip install flake8
       script: flake8 src/
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,9 @@ cache:
 
 python:
   - "3.6"
-
-matrix:
-  include:
-    - os: osx
-      language: generic
-    - os: linux
+os:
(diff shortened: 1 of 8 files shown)


FrickTobias:
Looks nice, but I'll wait with merging until the osx check is fixed.

marcelm:
The osx problem, as you may have seen, has nothing to do with the actual test, but is a download failure on the Travis side. This has occurred repeatedly over the last days. I’ve been restarting the Travis jobs to get the osx tests running.

To be honest, if this continues like this, I’m inclined to mark the osx tests as "allowed failures" in the future.

marcelm:
It appears the fact that it is sometimes *working* is actually the weird part. According to https://docs.travis-ci.com/user/languages/python/

> Python builds are not available on the macOS environment.

I unfortunately cannot assist with fixing this as I don’t have any macOS experience. I would like to set `allow_failures` appropriately until one of you has a chance to take a look.

Some of the comments I have found while googling indicate that it is possible to circumvent the problem, perhaps by setting `language: generic` instead of `language: python` (but then Python needs to be installed manually).

FrickTobias:
Interesting, I have not had this problem. Was it before or after you changed the `.travis.yml` file?

Currently we are (from what I gather) using that solution, but it is written as a matrix (I didn't get how to write it without it when I implemented it).

```
matrix:
  include:
    - os: osx
      language: generic
    - os: linux
```

FrickTobias:
I tested [branching](https://github.com/FrickTobias/BLR/tree/flake8_osx) where I changed `jobs` to include the previous version's solution for running `osx` with `language: generic` which seemingly solved the problem.  

```
jobs:
  include:
    - os: linux
    - os: osx
      language: generic
    - name: flake8
      before_install:
      install: python3 -m pip install flake8
      script: flake8 src/
```

marcelm:
Ok, I’ll add that to this PR

marcelm:
Hm, I think we can just set `language: generic` globally since we’re not using the Travis-installed Python anyway. Let me try that instead.

marcelm:
Cool, it’s working now. Thanks for investigating! This can now be merged.

FrickTobias:
No problem!

PR #105, opened 2019-09-26
Add an "init" subcommand

marcelm:
This allows one to create an analysis directory with `blr init --reads1=path/to/sample.R1.fastq.gz workdir`, which should be easier than manually creating the symlinks.

This will also rename the configuration file to `blr.yaml`.


src/blr/cli/init.py
```diff
@@ -0,0 +1,93 @@
...
+CONFIGURATION_FILE_NAME = "blr.yaml"
+
+
+def add_arguments(parser):
+    parser.add_argument(
+        "--reads1",
+        "--r1",
+        type=Path,
+        metavar="READS",
+        help="First paired-end read file (.fastq.gz). The second is found automatically.",
```
pontushojer:
```suggestion
        help="First paired-end read file (.fastq.gz). The second is found automatically.",
        required=True
```

Instead of having to check if `args.read1` is `None` in the main functions you could just set it as an required argument.

src/blr/cli/init.py
```diff
@@ -0,0 +1,93 @@
...
+    init(args.directory, args.reads1)
+
+
+def init(directory: Path, reads1: Path):
+    if " " in str(directory):
+        logger.error("The name of the analysis directory must not contain spaces")
+        sys.exit(1)
+
+    if directory.exists():
+        logger.error(f"The target directory {directory!r} already exists.")
```
pontushojer: Should not the script exit here if the directory exists? Otherwise it will just continue until it tries to create the directory in the `create_and_populate_analysis_directory` function and raise an error there.

marcelm: You’re right; the `sys.exit(1)` apparently got lost after I copied the code over from IgDiscover.

src/blr/cli/init.py
marcelm: Good point, I forgot argparse can do that.

marcelm:
I’ve updated the PR with the suggested changes.

src/blr/cli/init.py
```diff
@@ -0,0 +1,95 @@
...
+        required=True,
+        metavar="READS",
+        help="First paired-end read file (.fastq.gz). The second is found automatically.",
+    )
+    parser.add_argument("directory", type=Path, help="New analysis directory to create")
+
+
+def main(args):
+    if args.reads1 is None:
+        logger.error("--reads1 must be given")
```
pontushojer: With the change to the read1 argument we can remove this check as it does the same thing.

marcelm: done

FrickTobias:
Looks really nice, I just have two questions and two minor comments. My guess is you've written it like you have because of experience in dealing with users but I thought it best to ask.

#### Questions

What is the point of explicitly writing this as a `try` syntax, rather than just writing `directory.mkdir()` since the exception only seem to be returning the error message? This seem to be a recurring way of formulating it.
```
def create_and_populate_analysis_directory(directory: Path, reads1: Path, reads2: Path):
    try:
        directory.mkdir()
    except OSError as e:
        logger.error(e)
        sys.exit(1)
```

Is it important to force users to not add space to their analysis folder names? 
```
def init(directory: Path, reads1: Path):
    if " " in str(directory):
        logger.error("The name of the analysis directory must not contain spaces")
        sys.exit(1)
```

#### Minor comments

We're mixing f-strings and %s-strings, is there a particular reason?

Should we possibly include a description of the correct format since it currently is not described more than being automatic? I'm think we could describe the patter; identical strings with a 2 instead of a 1 somewhere in the name.
```
if reads2 is None:
    logger.error("Could not determine second file of paired-end reads")
    sys.exit(1)
```

marcelm:
> What is the point of explicitly writing this as a try syntax, rather than just writing directory.mkdir() since the exception only seem to be returning the error message?

(I just noticed that there is an explicit `if directory.exists()` check in addition to the `try: ... except:` block. I’ll need to remove the former.)

Regarding your question: I have two reasons, one is about the appearance and the other is a conceptual one.

First, the appearance. The version with `try` prints (depending on how the logger was configured) a simple `[Errno 17] File exists: 'tmp'`, which is concise and clear about where the problem is. The bare `.mkdir()` will result in a traceback, which is potentially long and confusing:
```
Traceback (most recent call last):
  File "example.py", line 18, in <module>
    main()
  File "a.py", line 16, in main
    a()
  File "a.py", line 4, in a
    b()
  File "a.py", line 7, in b
    c()
  File "a.py", line 10, in c
    d()
  File "a.py", line 13, in d
    os.mkdir("tmp")
FileExistsError: [Errno 17] File exists: 'tmp'
```
It also happens to contain the error message at the end, so many users will probably be ok with it, but there really is no need to tell the user where exactly in the program the error occurred. They just need to know what the problem is with the input they supplied.

One other problem is that you cannot change the last line: It will always say `FileExistsError: [Errno 17] File exists: ...` and has even changed between Python versions. The message is redundant and I could imagine one may want to change it so it says "File or directory with the same name already exists" or something like that, just to be a bit more user friendly (the concept that a directory *is* a file may not be obvious to a user). To do this change, catching the exception would be necessary anyway. (I realize this is not done in my code above – although it probably should –, but at least the message as printed will not contain the `FileExistsError:` part, which is really a Python-specific thing and redundant with the `File exists` text after it.)

The second reason is a conceptual one: If the program aborts due to a traceback, I consider this a bug.

Errors can come in various forms. All errors that I can anticipate, such as invalid command-line arguments, incorrectly formatted files, incorrect file permissions etc. should be dealt with within the application so I can take appropriate actions, such as to re-try to establish a network connection, to ask the user for a different file name, or, as in our case, to just print a helpful error message and exit with an error code so that the user can try again.

On the other hand, the remaining errors that I did not anticipate will throw a traceback. Tracebacks are meant for the programmer. They are intended to contain enough information to help one understand exactly where in the code the error occurred and what the problem was. With that, I can then either fix a bug in the program’s logic (perhaps a variable was initialized incorrectly) or, if the problem is actually with the input data, turn the error into an anticipated one by catching the exception and handling it appropriately.

I admit that Python lets us blur the distinction between "anticipated error" vs "bug" somewhat because printing the traceback and exiting the program is what we would have done anyway. Sometimes, in particular in small throwaway scripts, it’s totally fine to just let the traceback be printed and use that as the mechanism to report errors to users. For certain types of errors (such as MemoryError), there’s not much choice anyway. However, here we should absolutely do it because providing the name of an existing directory is an error we can easily anticipate.

> Is it important to force users to not add space to their analysis folder names?

We’ve been bitten by this in IgDiscover once (where we use the name of the analysis directory as sample name in some places), which is why I left it in. I think it would work in blr currently, but it’s easy to introduce changes that don’t take into account that the directory name can have spaces in it. If we want to allow this, we need to add a test for it. But to be honest, I don’t think it’s worth allowing spaces – it’s not a real limitation since people can always use underscores or dashes as separators instead.

> We're mixing f-strings and %s-strings, is there a particular reason?

Nah, the only reason is that I was almost done with the script until I noticed that we’re on Python 3.6 and can therefore actually use f-strings. I can convert the others as well, but let me explain why I’m still slightly torn between whether to use them or not.

The thing with `logger.xyz("message: %s %s", arg1, arg2)` is that formatting the strings is only done if the log level *xyz* is enabled. If the log level is disabled, the parameters are passed in but then discarded.

Passing an f-string would circumvent this optimization, so it goes a bit against the idea of the logging methods. However, I’ve recently come to the conclusion that the convenience of being able to use f-strings outweighs this ("practicality beats purity") and that a performance difference is nonexistent anyway for those messages that *are* printed. So I’ve started to use f-strings even here, but – as you can see – I haven’t been converted old occurrences.

FrickTobias:
Alright, sounds reasonable! Thanks for the elaborate (and well-written) explanation!

PR #109, opened 2019-10-02
Starcode for barcode clustering 

pontushojer:
I have now implemented [starcode](https://github.com/gui11aume/starcode) for barcode clustering. Starcode clusters sequences within a given Levenshtein distance from each other if the count ratios are within a certain threshold and this is also done recursively. We might want to test different parameters for this in the future but for now I am using a edit distance of 2 and ration of 5 (default). 

I tested starcode on a full dataset (*unfortunately my notes are on uppmax which is down for maintenance at the moment so I cannot post the specifics right now*). The dataset was barcode from a SM10 run with about 40Gb data. I tested runtime in correlation to edit distance which something like this:

| Edit distance (-d) | Runtime |
| --- | ---|
| 1 | ~1 h |
| 2 | ~1.5 h |
| 3 | 4 h? |
| 4 | Much longer?? |

This is clearly faster than cd-hit, at least for lower edit distances, ,which in comparison took about 30 min for 1/64 of the data (one nucleotide indexed file TTT.fa from the same dataset). This was also able to handle the full dataset without memory issues. 

Because of this, and in discussion with Tobias, I also removed cd-hit from the pipeline. 

Other changes: 
- New script for correcting the barcode sequences and tagging the FASTQ files before mapping. 
- Changed trimming to separate barcode extraction from read trimming. This is to extract barcodes more accurately and include possible INDELs in the sequence. This will however increase runtime slightly as there is no way, that I have found, to perform these step together using cutadapt. Perhaps @marcelm has some ideas about this?

*Note the PR is currently failing due to that the test files cannot be downloaded from UPPMAX due to maintenance. *


marcelm:
I’d appreciate if I could get a chance to review this before it gets merged. I don’t have time today, though.

pontushojer:
> I’d appreciate if I could get a chance to review this before it gets merged. I don’t have time today, though.

No stress! I will leave this open, I also want to add the real figures for the starcode test run.

src/blr/rules/trim.smk
```diff
@@ -1,48 +1,70 @@
...
+              Presence depencd on
+              insert length
 """
 
+h1="CAGTTGATCATCAGCAGGTAATCTGG"
+DBS="N"*20
+h2="CATGACCTCTTGGAACTGTCAGATGTGTATAAGAGACAG"
+h3="CTGTCTCTTATACACATCT"
+trim_len=sum(map(len, [h1,DBS,h2]))
+extract_len=len(h1)
```
FrickTobias: Maybe these should be written to a config file? Since we've already changed a lot would could also postpone it to a a future PR/issue.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    # If no output1 is given output is sent to stdout
+    if not args.output1:
+        logger.info(f"Writing output to stdout.")
+        args.output1 = sys.stdout.buffer
+        args.output2 = None
+
+    out_interleaved = True if not args.output2 else False
+    logger.info(f"Output detected as {'interleaved fastq.' if out_interleaved else 'paired fastq.'}")
+
+    reads_skipped = list()
```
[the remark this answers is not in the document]
pontushojer: Acctually this will probably occur a bit as we have separated the trimming on the reads from the barcode extraction. Then are currently having some small differences in output that i think might persist. Thus I think it is good to log them somewhere, possibly we could also just write them to a separate file with unknown reads.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    corrected canonical sequence
+    :param open_file: starcode tabular output file.
+    :return: dict: raw sequences pointing to a nametuple containe corrected canonical sequence and id.
+    """
+    target = namedtuple("Cluster", ['seq', 'id'])
+    corrected_barcodes = dict()
+    for id, cluster in tqdm(enumerate(open_file.readlines(), start=1), desc="Corrected barcodes processed"):
+        canonical_seq, _, cluster_seqs = cluster.strip().split("\t", maxsplit=3)
+        cluster_target = target(seq=canonical_seq, id=id)
+        corrected_barcodes.update({raw_seq: cluster_target for raw_seq in cluster_seqs.split(",")})
```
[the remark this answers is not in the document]
pontushojer: Yes but we need that as we have to tag the reads header with the corrected barcode sequence, which for some reads is the same as the raw sequence. I could include a condition that checks if the read barcode matches a canonical sequences but that would just make it more complicated in opinion.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+        "-o2", "--output2", default=None, metavar='<OUTPUT2>',
+        help="Output FASTQ/FASTA name for read2. If not specified but -o1 given the result is written"
+             " as interleaved to o1.")
+    parser.add_argument(
+        "-rb", "--barcodes-raw", required=True, metavar='<RAW BARCODES>',
+        help="FASTQ/FASTA for raw barcodes.")
+    parser.add_argument(
+        "-cb", "--barcodes-corrected", required=True, metavar='<CORRECTED BARCODES>',
+        help="FASTQ/FASTA for error corrected barcodes. Currently accepts output from starcode clustering with "
+             "'--print-clusters' enabled.")
```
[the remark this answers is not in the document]
pontushojer: Yes was thinking about this but then I got stuck on how to handle if there is a interleaved file as input. But thinking a bit more about this I can just switch the order of these arguments and there should be no problem.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def parse_corrected_barcodes(open_file):
+    """
+    Parse starcode cluster output and return a dictionary with raw sequences pointing to a
+    corrected canonical sequence
+    :param open_file: starcode tabular output file.
+    :return: dict: raw sequences pointing to a nametuple containe corrected canonical sequence and id.
+    """
+    target = namedtuple("Cluster", ['seq', 'id'])
```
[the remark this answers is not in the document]
pontushojer: > It took me a little while to figure this out - since we're dealing with both reads names and cluster integers here, `id` was a little big ambiguous. Maybe we could substitute it with `fastq_header` or something? I am mainly thinking about it being easier to understand in the future.

I can maybe change it to `index`for now as I think this better describes it. The only reason this is included is that we have been using these cluster integer numbers to represent the "canonical sequence" that is given by starcode, idealy I think we should move away from this integer entirely and just use the canonical sequence directly. This might however interfere with some thing we have written earlier so we have to check how to do the change. Also, to use the barcode sequence is standard for most 10x or other linked-read related tools I have seen so this would standardize our process a bit more

pontushojer:
**TEST RESULTS**
Dataset: XVII.SM10.reseq_2
Size: 29 Gb (R1) + 40 Gb (R2)
Starcode version: starcode-v1.3 17-07-2018

command: 
```
starcode --print-clusters -t 20 -d <VARIED> -o output -i input
```

**Test 1: Starcode on barcode FASTA**
File size: 24Gb

Edit dist (-d) | User time
-- | --
1 | 94m16.737s
2 | 134m7.640s

**Test 2: Starcode on aggregated barcodes file**
File format: BARCODE_SEQUENCE\tCOUNT
File size: 1.3Gb

Edit dist (-d) | User time
-- | --
1 | 61m45.917s
2 | 102m21.404s
3 | 486m30.105s

As you see the time to process one file increases by ~30 min if you use the FASTA format as in test 1 instead of the starcode aggregated format used in test 2. This however requires you to sort and aggregate the sequences from the FASTA/FASTQ which takes about 80 min for me using this primitive oneliner:

```
cat $file | paste - - | cut -f 2 | sort | uniq -c | awk '{print $2,$1}' > $file"sort.sc"
```

Maybe there is a faster way to do this but I think that as starcode handles this quite fast we can use input FASTA/FASTQ directly.

src/blr/cli/tagfastq.py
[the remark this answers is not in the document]
pontushojer: >I think it would make more sense to keep a count on how many reads are without barcode, or if you really find it useful, at least make it a set rather than a list for memory efficiency.

FYI [sets are more memory consuming than lists. ](https://stackoverflow.com/questions/13547883/is-python-set-more-space-efficient-than-list)

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    # Get the corrected barcodes and create a dictionary pointing each raw barcode to its canonical sequence.
+
+    with open(args.barcodes_corrected, "r") as reader:
+        corrected_barcodes = parse_corrected_barcodes(reader)
+
+    # Get the raw barcodes and create a dictionary pointing each read header to the raw barcode sequence.
+    with dnaio.open(args.barcodes_raw, mode="r") as reader:
+        header_barcodes = parser_raw_barcodes(reader)
+
+    in_interleaved = True if not args.input2 else False
```
marcelm: Should be:
```
in_interleaved = not args.input2
```

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    logger.info(f"Input detected as {'interleaved fastq.' if in_interleaved else 'paired fastq.'}")
+
+    # If no output1 is given output is sent to stdout
+    if not args.output1:
+        logger.info(f"Writing output to stdout.")
```
marcelm: This doesn’t have to be an f-string

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+        args.output1 = sys.stdout.buffer
+        args.output2 = None
+
+    out_interleaved = True if not args.output2 else False
```
marcelm: `not args.output2`

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    logger.info(f"Output detected as {'interleaved fastq.' if out_interleaved else 'paired fastq.'}")
```
marcelm: * fastq → FASTQ (in user-visible strings)
* move the FASTQ out of the conditional since it is common to both strings

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    target = namedtuple("Cluster", ['seq', 'id'])
+    corrected_barcodes = dict()
+    for id, cluster in tqdm(enumerate(open_file.readlines(), start=1), desc="Corrected barcodes processed"):
+        canonical_seq, _, cluster_seqs = cluster.strip().split("\t", maxsplit=3)
+        cluster_target = target(seq=canonical_seq, id=id)
+        corrected_barcodes.update({raw_seq: cluster_target for raw_seq in cluster_seqs.split(",")})
+    return corrected_barcodes
+
+
+def parser_raw_barcodes(open_file):
```
marcelm: Should be named `parse_raw_barcodes` (verb, no "r")

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    Parse FASTA/FASTQ containing barcode sequences and return a dictionary with entry headers pointing to a
```
marcelm: This and other lines are too long. I need to scroll in the GitHub review interface.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    header_barcodes = dict()
+    for barcode in tqdm(open_file, desc="Raw barcodes processed"):
+        header, _ = barcode.name.split(maxsplit=1)
+        header_barcodes[header] = barcode.sequence
+    return header_barcodes
+
+
+def add_arguments(parser):
+    parser.add_argument(
+        "input1", metavar='<INPUT1>',
```
marcelm: Please leave out the `metavar=` here and in all places, where only the angled brackets are added. They don’t make the help output more readable.

In general, if you want behavior different from the argparse default, you should change how argparse constructs the help string, not adjust the string in each call to `add_argument`.

src/blr/cli/tagfastq.py
FrickTobias: Sweet syntax 🥇

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+        help="Input FASTQ/FASTA file. Assumes to contain read1 if given with second input file. If only "
+             "input1 is given, input is assumed to be an interleaved. If reading from stdin"
+             "is requested use '-' as a placeholder.")
+    parser.add_argument(
+        "input2", nargs='?', metavar='<INPUT2>',
+        help="Input  FASTQ/FASTA for read2 for paired-end read. Leave empty if using interleaved.")
+    parser.add_argument(
+        "-o1", "--output1", default=None, metavar='<OUTPUT1>',
```
marcelm: Single dashes should generally only be used with single-character command-line options. (This is a convention established by `getopt_long`.) The reason is that you can merge multiple single-character options that don’t require arguments into one: `ls -la` instead of `ls -l -a`. This gets ambiguous as soon as you allow multi-character options with a single dash. argparse somehow handles this, but I would not rely on it. Cutadapt uses `-o` and `-p` because of this. You could also use `--o1` instead.

Same comment applies to options below.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+        "-rb", "--barcodes-raw", required=True, metavar='<RAW BARCODES>',
+        help="FASTQ/FASTA for raw barcodes.")
+    parser.add_argument(
+        "-cb", "--barcodes-corrected", required=True, metavar='<CORRECTED BARCODES>',
+        help="FASTQ/FASTA for error corrected barcodes. Currently accepts output from starcode clustering with "
+             "'--print-clusters' enabled.")
+    parser.add_argument(
+        "-bc", "--barcode-cluster-tag", metavar="<STRING>", type=str, default="BX",
+        help="BAM file tag where barcode cluster id is stored. 10x genomics longranger output "
+             "uses 'BX' for their error corrected barcodes. DEFAULT: BX")
```
marcelm: Don’t hard-code the default in the help message, but use `%(default)s` instead.

src/blr/config.schema.yaml
```diff
@@ -1,10 +1,6 @@
 $schema: "http://json-schema.org/draft-06/schema#"
 description: Configuration parameters for running BLR pipeline
 properties:
-  index_nucleotides:
-    type: integer
-    description: Number of index nucleotide to use when splitting for barcode clustering.
-    default: 3
```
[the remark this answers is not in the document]
marcelm: Doesn’t the PR do away with the extra splitting step entirely?

src/blr/cli/tagfastq.py
FrickTobias: In line with your previous comment about global configuration for argparse, maybe we could make argparse always add a `DEFAULT: %(default)s` string to help messages whenever `default`is not `None`?

src/blr/rules/trim.smk
```diff
@@ -1,48 +1,70 @@
...
+
+READ 2 LAYOUT
+
+5'-NNNN...NNNN(CTGTCTCTTATACACATCT)-3'
+   <---gDNA--><---------h3-------->
+              Presence depencd on
+              insert length
 """
 
+h1="CAGTTGATCATCAGCAGGTAATCTGG"
```
marcelm: Please add spaces around operators

src/blr/rules/trim.smk
```diff
@@ -1,48 +1,70 @@
...
-    log:
-        a = "trimmed-a.log",
-        b = "trimmed-b.log"
+        r2_fastq="reads.2.fastq.gz",
+    log: "cutadapt_trim.log"
     threads: 20
     shell:
-        "cutadapt"
-        " -g 'XNNNCAGTTGATCATCAGCAGGTAATCTGG;min_overlap=26'"
+        "cutadapt" #Initial trim
```
marcelm: The `inital trim` comment doesn’t add anything, I suggest to leave it out.

src/blr/cli/tagbam.py
```diff
@@ -16,65 +14,23 @@ def main(args):
...
-        else:
-            cluster_sequence = pattern.search(line).group(1)
-            cluster_dict[cluster_sequence] = cluster_id
-
-    return cluster_dict
-
-
 def add_arguments(parser):
     parser.add_argument("input_mapped_bam", metavar="<INPUT_BAM>",
                         help=".bam file with mapped reads which is to be tagged with barcode id:s.")
```
marcelm: "id:s" is a Swedish plural ;-). Use "ids" or "IDs".

marcelm: It should be `BAM file`.

src/blr/cli/tagfastq.py
```diff
@@ -0,0 +1,132 @@
...
+    reads_skipped = list()
+    # Parse input FASTA/FASTQ for read1 and read2 and write output
+    with dnaio.open(args.input1, file2=args.input2, interleaved=in_interleaved, mode="r") as reader, \
+            dnaio.open(args.output1, file2=args.output2, interleaved=out_interleaved, mode="w") as writer:
+        for read1, read2 in tqdm(reader, desc='Read pairs processed'):
+            # Header parsing
+            name_and_pos_r1, read_and_index_r1 = read1.name.split(maxsplit=1)
+            name_and_pos_r2, read_and_index_r2 = read2.name.split(maxsplit=1)
+
+            assert name_and_pos_r1 == name_and_pos_r2
```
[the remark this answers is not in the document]
marcelm: Yes, dnaio checks whether the files are correctly paired. Among other checks, it checks whether the read names match.

src/blr/cli/tagfastq.py
marcelm: Sounds good in principle but there are too many exceptions when this wouldn’t work. I think this is something that needs to decided from option to option since it’s meant as human-readable text. For conciseness, for example, I sometimes do not write out the default if it is clear from the context. For example, if `--reverse` means "Print foo file in reverse", it’s obvious that the default is to not print the file in reverse. And whether I have encode this as `None` and `True` or `forward` and `reverse` in the value is independent of this.

(By the awy, I suggest to use normal casing: `Default: %(default)s`.)

FrickTobias: Though I guess that specific case would be solved by having a convention to use the `"store_true"` flag when applicable (for all "binary" options). Either way since this was just one example I'll trust your experience regarding this and leave it up for future me to revisit this if I feel it is necessary.

FrickTobias:
There is a lot of feedback from here which is applicable to numerous submodules, I'll create issues listing them.

src/blr/rules/trim.smk
marcelm: This is a really helpful ASCII drawing. Can you please also write what h1 to h3 and DBS mean? (I can guess, but I shouldn’t have to.) This could perhaps also be moved to the documentation at some point.

src/blr/cli/tagfastq.py
marcelm: I don’t want to argue from authority, so I’ve looked through the options in Cutadapt. It would work there for seven of the options if you extend the rules to not print the default when `default=''` or `default=[]` (for `action="append"`). On the other hand, there are a couple of options for which I print out a default even though `default=None`, such as "Default: no limit" (for maximum read length) or "Default: output to same file as trimmed reads" (for `--untrimmed-output`).

If at some point this is the only thing in the code that prevents it from being beautiful, I’d so go for it.

FrickTobias: Thanks for taking the time to look it up! 

Yeah, that makes sense that some options could require a more readable or non-programming term to explain defaults to a user.

marcelm:
Regarding the extra barcode extraction step, I don’t think avoiding this is easily possible with Cutadapt at the moment. Also, I don’t have a good idea for a general syntax for this on the command-line, so I would even know how to start.

However, if the extra runtime becomes sufficiently annoying, I can look into using Cutadapt directly as a Python module. It doesn’t have an official API, but if we pin the version, this shouldn’t be a problem.

src/blr/rules/trim.smk
```diff
@@ -1,48 +1,70 @@
...
-        " -o1 - |" 
-        "cutadapt"
-        " -a ^CATGACCTCTTGGAACTGTCAGATGTGTATAAGAGACAG...CTGTCTCTTATACACATCT"
-        " -A CTGTCTCTTATACACATCT"
+        "> {log} "
+
+rule extract_DBS:
+    # Extract barcode sequence from read1 FASTQ
+    output:
+        fastq="barcodes.fastq",
```
marcelm: Are the quality values actually needed here? If not, you could save some space (and I/O) by writing a `barcodes.fasta` file instead. Just change the name of the output file and Cutadapt will convert the format.

FrickTobias: From the starcode paper "raw sequence with count, FASTA or FASTQ (in which case starcode ignores the quality).", so fasta should be sufficient.

pontushojer: I will change

src/blr/rules/trim.smk
pontushojer: Since I moved these to the config file I added a small comment about them. Hopefully they should be easier to understand. I agree that the drawing could be useful in the docs, I included it here just to keep my own sanity while doing changes :)

src/blr/cli/tagfastq.py
pontushojer: > Regarding the extra barcode extraction step, I don’t think avoiding this is easily possible with Cutadapt at the moment. Also, I don’t have a good idea for a general syntax for this on the command-line, so I would even know how to start.
> 
> However, if the extra runtime becomes sufficiently annoying, I can look into using Cutadapt directly as a Python module. It doesn’t have an official API, but if we pin the version, this shouldn’t be a problem.

Then we will keep this for now and change it only if it becomes necessary. 

I have include (hopefully) all changes now. Thank you for your thorough reviews!!

pontushojer:
> Regarding the extra barcode extraction step, I don’t think avoiding this is easily possible with Cutadapt at the moment. Also, I don’t have a good idea for a general syntax for this on the command-line, so I would even know how to start.
> 
> However, if the extra runtime becomes sufficiently annoying, I can look into using Cutadapt directly as a Python module. It doesn’t have an official API, but if we pin the version, this shouldn’t be a problem.

Then we will keep this for now and change it only if it becomes necessary. 

I have include (hopefully) all changes now. Thank you for your thorough reviews!!

PR #111, opened 2019-10-07
Allow to choose BWA-MEM instead of Bowtie2 as read mapper

marcelm:
This introduces a new configuration setting named `read_mapper` and
also renames the `bowtie2_reference` setting to `genome_reference`.

Run times when mapping 20 million read pairs (repeated three times):

* Bowtie2: 43m 33s, 43m 27s, 43m 22s
* BWA-MEM: 33m 23s, 33m 30s, 33m 31s

Looking at these numbers, we should probably change the default mapper, but I haven’t done so in this PR.

I’ve also added a commit that removes usage of `params` when the only purpose is to assign `config["key"]` to a variable which can the be used within a shell command. In those cases, `{config[key]}` can be used directly instead.

*Note*: This will probably need to be rebased once #109 is merged.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -14,7 +14,6 @@ blr --version
 rm -rf outdir-bwa
 blr init --r1=testdata/reads.1.fastq.gz outdir-bwa
 sed 's|read_mapper: .*|read_mapper: bwa|' tests/test_config.yaml > outdir-bwa/blr.yaml
-#echo "genome_reference:
 
 pushd outdir-bwa
 blr run
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -9,14 +9,26 @@ cd-hit --help | head -n 1 || true
 snakemake --version
 blr --version
 
-( cd testdata && bowtie2-build chr1mini.fasta chr1mini > /dev/null )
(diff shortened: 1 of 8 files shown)


tests/run.sh
```diff
@@ -9,14 +9,26 @@ cd-hit --help | head -n 1 || true
...
+( cd testdata && bwa index chr1mini.fasta )
 
-rm -rf outdir
-blr init --r1=testdata/reads.1.fastq.gz outdir
-cp tests/test_config.yaml outdir/blr.yaml
-cd outdir
+rm -rf outdir-bwa
+blr init --r1=testdata/reads.1.fastq.gz outdir-bwa
+sed 's|read_mapper: .*|read_mapper: bwa|' tests/test_config.yaml > outdir-bwa/blr.yaml
+#echo "genome_reference:
```
pontushojer:
```suggestion
```

I guess this should be removed?

marcelm: Yes.

FrickTobias:
Looks good and I agree we should change mapper. It's worth to notice bwa can also tag reads with the barcode from the header meaning we could skip that step later as well. However it might be best to wait until Emerald is investigated to do any such changes.

Issue #110, opened 2019-10-04
Consistency in argparse and comments


FrickTobias:
In line with what @marcelm pointed out in #109 we should also apply some of these comments in other scripts whenever applicable. The ones I have found we currently do not follow are:

**Argparser**
- help: `%(default)s` instead of hardcoded strings
- help: Write BAM rather than .bam
- Omit `type=str` in  (it is default)
- Use double dash `--` for multi-character option names (not only their long names)

**Comments**
- Write abbreviations without colon (e.g. ID:s => IDs)

pontushojer:
This was fixed in #118

PR #118, opened 2019-10-08
Refactoring arguments in python scripts

pontushojer:
See #110. 

Added documentation for which SAM tags are used and under which options flag they should be implemented in python scripts, along with a description for each. 
  
Removed `extractbarcodes.py` which I forgot to remove in PR #109.


FrickTobias:
If you've only changed the documentation, help messages and removed a script - i.e. not changed the analysis - why has the md5sum changed?

pontushojer:
Also tagged the trimmed fastq files with the original barcode sequence. Hence the md5sum update.

Issue #121, opened 2019-10-11
Tag reads with barcode sequence


pontushojer:
Currently reads are tagged with barcode cluster ids. This comes from when we used cd-hit and got a cluster index rather than a canonical sequence for the cluster. Using the barcode includes more valuable information and is also more similar to the longranger format (like it or not this will probably be the basis of manny tools we want to apply). Possibly we also want to include a GEM group there also.

PR #123, opened 2019-10-18
Updated barcode information handling 

pontushojer:
Updated barcode information handling in FASTQ and BAM files so that barcode strings can be handled. This is how barcodes are handled in 10x genomics data and contain more useful information than the integer representing the `cluster id` previously used. I also implemented some changes to handling of alignment/read headers and SAM tags. 

- `tagfastq` now tags reads with the corrected and raw barcode as SAM tags separated by undescore '_' e.g. `@ST:E00269:339:H27G2CCX2:7:1608_RX:Z:GGCGTAACTACGTTAGCTCG_BX:Z:GGCGTAACTACGTTAGCTCG 1:N:0:NGGCAGAA`. Reads without barcodes appear without these tags.
- `tagbam` transfers the tag information from the header to the designated SAM tags. Note that this now removed the info from the header.
- `clusterrmdup` now handles strings instead of integers. The sort function is used as previously to direct all duplicate connected clusters to the same final cluster.
- `filterclusters` now only appends "_FILTERED" to reads filtered out and removes the barcode tags rather than replacing them with the string "FILTERED". This is because the string "FILTERED" could be interpreted as a barcode in downstream analysis. Thus it feels safe to just remove it.
- Updated md5sum based on edits.

Issue #121

### DIFF
--- a/environment.linux.lock.yml
+++ b/environment.linux.lock.yml
@@ -6,56 +6,58 @@ dependencies:
   - _libgcc_mutex=0.1=main
   - alsa-lib=1.1.5=h516909a_1001
   - appdirs=1.4.3=py_1
-  - asn1crypto=0.24.0=py36_1003
-  - attrs=19.2.0=py_0
-  - blas=2.12=openblas
+  - asn1crypto=1.2.0=py36_0
+  - attrs=19.3.0=py_0
+  - blas=2.14=openblas
   - bowtie2=2.3.5=py36he860b03_0
   - bwa=0.7.17=hed695b0_6
   - bzip2=1.0.8=h516909a_1
   - ca-certificates=2019.9.11=hecc5488_0
   - certifi=2019.9.11=py36_0
-  - cffi=1.12.3=py36h8022711_0
+  - cffi=1.13.1=py36h8022711_0
   - chardet=3.0.4=py36_1003
   - configargparse=0.13.0=py_1
   - cryptography=2.7=py36h72c5cf5_0
   - curl=7.65.3=hf8cf82a_0
   - cutadapt=2.5=py36h516909a_0
   - datrie=0.8=py36h516909a_0
-  - dnaio=0.3=py36h14c3975_1
+  - dnaio=0.4=py36h516909a_0
   - docutils=0.15.2=py36_0
   - fontconfig=2.13.1=h86ecdb6_1001
   - freetype=2.10.0=he983fc9_1
   - giflib=5.1.7=h516909a_1
   - gitdb2=2.0.6=py_0
-  - gitpython=3.0.2=py_0
+  - gitpython=3.0.4=py_0
   - htslib=1.9=ha228f0b_7
   - icu=64.2=he1b5a44_1
   - idna=2.8=py36_1000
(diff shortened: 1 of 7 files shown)


FrickTobias:
Nice!

Could we also take this opportunity to go from an underline separator to a separator? This would make it compatible with the bwa `-C` flag which would mean we don't need to tag the bamfile, as it should include all tags found in the header which follow a SAM format if I understand it correctly.

From [bwa documention](http://bio-bwa.sourceforge.net/bwa.shtml):

-C | Append append FASTA/Q comment to SAM output. This option can be used to transfer read meta information (e.g. barcode) to the SAM output. Note that the FASTA/Q comment (the string after a space in the header line) must conform the SAM spec (e.g. BC:Z:CGTAC). Malformated comments lead to incorrect SAM output.
-- | --

pontushojer:
> Nice!
> 
> Could we also take this opportunity to go from an underline separator to a separator? This would make it compatible with the bwa `-C` flag which would mean we don't need to tag the bamfile, as it should include all tags found in the header which follow a SAM format if I understand it correctly.
> 
> From [bwa documention](http://bio-bwa.sourceforge.net/bwa.shtml):
> 
> -C	Append append FASTA/Q comment to SAM output. This option can be used to transfer read meta information (e.g. barcode) to the SAM output. Note that the FASTA/Q comment (the string after a space in the header line) must conform the SAM spec (e.g. BC:Z:CGTAC). Malformated comments lead to incorrect SAM output.

Sweet! That is a nice feature. We would however need it for mapping with bowtie2, maybe we could have the separator as an argument option e.g -s, --separator.

pontushojer:
@FrickTobias So will leave out running the bwa with the `-c`option for  now. I did some testing and it seams like it only works for the first entry in the header i.e. for this header

```
@ST-E00269:339:H27G2CCX2:7:1102:25175:1608 RX:Z:GGCGTAACTACGTTAGCTCG BX:Z:GGCGTAACTACGTTAGCTCG 1:N:0:NGGCAGAA
```
there will only be one tag "RG" with the value "GGCGTAACTACGTTAGCTCG BX:Z:GGCGTAACTACGTTAGCTCG 1:N:0:NGGCAGAA" which is not what we want. I could rewrite it so that only the corrected barcode is in the header and then it would work but then we would loose the information about the raw barcodes sequence (and the sequencing index but that maybe not important). 

Anyway I added a option to change the separator should we need it.

marcelm:
The docs that say that "the FASTA/Q comment [..] must conform the SAM spec" mean this literally: Anything after the first space is copied verbatim into the output SAM. This includes any spaces, but SAM needs tab characters. If you make the read look like this, I believe it would work:
```
@ST-E00269:339:H27G2CCX2:7:1102:25175:1608_1:N:0:NGGCAGAA RX:Z:GGCGTAACTACGTTAGCTCG<TAB>BX:Z:GGCGTAACTACGTTAGCTCG 
```
Where I’ve used `<TAB>` to indicate a tab character.

Anyway, I would also argue that this relies on BWA being the read mapper too much and wouldn’t actually use this as long as we allow other mappers.

Issue #119, opened 2019-10-08
Print DAG functionality


pontushojer:
Enable the ability to print out the DAG flowchart for the Snakefile. We could do something similar to how I did in the [DBS-Pro](https://github.com/FrickTobias/DBS-Pro) `run`script.

PR #124, opened 2019-10-18
Print DAG for workflow

pontushojer:
Added the possibility to print a pdf with the DAG of the current workflow (issue #119).


FrickTobias:
I'll merge it subsequent to the prior PR.

PR #126, opened 2019-10-24
Start testing with pytest

marcelm:
The `tests/run.sh` script is becoming a bit crowded and hard to read because more and more tests are getting added to it. We should start using a proper test framework. This PR adds a couple of pytest tests that test the pipeline from Python. The idea would be to migrate all tests from the bash script eventually.

You’ll need to `conda install pytest` manually to run the tests locally.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,30 +1,58 @@
 from pathlib import Path
+import pysam
+import pytest
+from xopen import xopen
 
 from blr.cli.init import init
 from blr.cli.run import run
 
+TESTDATA_READS = Path("testdata/reads.1.fastq.gz")
 
-def copy_config(source, target, genome_reference=None):
-    """Copy config, possibly changing the genome_reference"""
+
+def count_bam_alignments(path):
+    with pysam.AlignmentFile(path) as af:
+        n = 0
+        for _ in af:
+            n += 1
+    return n
+
+
+def count_fastq_reads(path):
+    with xopen(path) as f:
+        n = 0
+        for _ in f:
+            n += 1
+    return n // 4
+
+
+def copy_config(source, target, genome_reference=None, read_mapper=None):
+    """Copy config, possibly changing genome_reference or read_mapper"""
 
     with open(source) as infile:
         with open(target, "w") as outfile:
             for line in infile:
                 if genome_reference is not None and line.startswith("genome_reference:"):
-                    line = "genome_reference: " + str(Path("testdata/chr1mini.fasta").absolute())
+                    path = Path("testdata/chr1mini.fasta").absolute()
+                    line = f"genome_reference: {path}\n"
+                if read_mapper is not None and line.startswith("read_mapper:"):
+                    line = f"read_mapper: {read_mapper}\n"
                 outfile.write(line)
 
 
 def test_init(tmpdir):
-    init(tmpdir / "analysis", Path("testdata/reads.1.fastq.gz"))
+    init(tmpdir / "analysis", TESTDATA_READS)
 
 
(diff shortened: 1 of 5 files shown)


marcelm:
In case you haven’t worked with pytest before, this PR serves as a way to show how it works: Just run `pytest tests/` and pytest will find and run the tests in the `tests/` directory automatically. You just need to make sure to put them into files that are named `test_...py` and that the tests themselves are functions named `test_...`.

A very nice pytest feature are the so-called fixtures, which define something that needs to run before and after a test (setup and teardown). In this PR, I’m using the `tmpdir` fixture, which will create a temporary directory before the test is run and delete it afterwards. Each test gets a new temporary directory in order to isolate tests as much as possible. If a test fails, the temporary directory is not removed so you can debug what went wrong. To specify that a test function uses a certain fixture, a parameter needs to be added to the function that has the same name as the fixture. pytest inspects the parameter names and calls the appropriate fixture.

Another helpful feature is the `parametrize` decorator. Applied to a test function, it will make pytest run the test once for each of the given arguments. You see an example in this PR.

tests/test_cli.py
```diff
@@ -0,0 +1,58 @@
...
+            n += 1
+    return n
+
+
+def count_fastq_reads(path):
+    with xopen(path) as f:
+        n = 0
+        for _ in f:
+            n += 1
+    return n // 4
```
FrickTobias: This expression will floor the result to the closest integer. Maybe we could also include a check so we know it has a root four? I'm thinking 
```suggestion
    return n // 4 if (n % 4) == 0 else error
```

```
>>> 10 // 4
2
>>> 11 // 4
2
>>> 12 // 4
3
```

FrickTobias: (but something meaningful instead of where I just wrote ´error´)

marcelm: I don’t think it’s too important to check that here because the FASTQ file will be run the pipeline anyway where other tools will check it. It was just a way to avoid hardcoding the number of reads in the test data FASTQ. But I agree, it doesn’t hurt to do a check here and there’s a risk someone might take the function and use it somewhere else where the input is not known beforehand.

I think I’ll just parse the file with `dnaio` here, which will even check whether it’s a correctly formatted FASTQ file.

marcelm:
> do you know why they've used @ in the beginning of the function name for @pytest.mark.parametrize? I've never seen it before.

The `@` indicates that `pytest.mark.parametrize` is a decorator.

A decorator is used to change the behavior of a function without changing
the function itself.

Here is an example usage:
```
@mydecorator
def myfunction():
    print("Hello")
```
One would say that the `myfunction` function is decorated with the `mydecorator` decorator.

This is essentially just a shortcut for this:
```
def myfunction():
    print("Hello")

myfunction = mydecorator(myfunction)
```
That is, the decorator is a function that returns a new function. The new
function then overwrites the original function.

Here is an example of how such a decorator could look. This one changes the
function so that it will print some text, then run the original function,
then print some more text:
```
def mydecorator(f):

    def newfunction():
        print("Before")
        f()
        print("After")

    return newfunction
```
With this, the decorated version of `myfunction()` will now print:
```
Before
Hello
After
```
Decorators can also be written in such a way that they accept parameters (as `pytest.mark.parametrize` does). And I should mention that classes can also be decorated (then the decorator just returns a new, modified class instead of a function).

tests/test_cli.py
FrickTobias: Yeah that sounds like a good idea, that means it will be less complex as well since we're not introducing several kinds of error handling for the same thing.

FrickTobias:
Alright! Thanks!

Issue #122, opened 2019-10-17
Simplify how binaries are found


marcelm:
The pipeline currently requires that the paths to Picard and HapCUT2 are set in the configuration file. This should not be necessary. Instead, we should simply expect that the programs are installed and available on the $PATH. Since we recommend users to install BLR with Conda, this will work.

This affects the `picard_command` and `hapcut2` configuration settings. For HapCUT2, we may need a Conda recipe first (#90).

FrickTobias:
Finished
- [x] picard
- [ ] HapCUT2

PR #127, opened 2019-10-25
Remove the picard_command configuration setting

marcelm:
We instead rely on $PATH for finding the binary.

See #122


FrickTobias:
Had to restart the travis run which failed due to timeout but now it works. I'll go ahead and merge when prior PR are finished.

PR #128, opened 2019-10-25
Avoid reading the entire .clstr file into memory

marcelm:
.readlines() reads all lines and returns a list. Instead, iterate over the
file object directly, which will only keep the current line in memory.


FrickTobias:
Nice catch! I'll go ahead and merge when prior PR are finished.

Issue #113, opened 2019-10-07
Test alternative duplicate marking programs


marcelm:
Consider whether possibly faster alternatives to Picard MarkDuplicates can be used:
* sambamba markdup
* samblaster

pontushojer:
**Run test on different duplicate marking programs.**

Dataset: XVII.SM10.reseq_2 mapped using bowtie2 and chr1 extracted. Final number of mapped alignments is 60,753,598 st.

Program versions:
- Picard: 2.20.4
- sambamba:  0.6.6
- samblaster: 0.1.24



**Test 1.** Threads = 20

Program | Real time | User time | %duplicates 
-- | -- | -- | --
Picard MarkDuplicates | 20m57.394s | 58m27.382s | 67.534
sambamba markdup | 3m9.613s | 42m14.153s | 67.534
samblaster | 2m6.075s | 0m59.907s | 64.64

**Test 2.** Threads = 2

Program | Real time | User time | %duplicates 
-- | -- | -- | --
Picard MarkDuplicates | 21m27.768s | 21m14.566s | 67.534 
sambamba markdup | 26m37.955s | 34m7.832s | 67.534
samblaster | 5m7.100s | 1m25.233s | 64.64

**Conclusion**
- samblaster is faster than both sambamba and picard on both 2 and 20 threads. 
- picard does not benefit from running in parallel unlike sambamba and samblaster.
- samblaster seems to be less sensitive in detecting duplicates (64.64% detected relative to 67.53%). 

Finally in testing I also want to point out a difference in implementation when using samblaster. Unlike picard and sambamba, samblaster requires a name-sorted SAM file for input and outputs in SAM format. This means that it should be implemented to take stdin directly from the mapper, for example:

```
bowtie2 -1 r1.fq -2 r2.fq -r ref.fa | samblaster | samtools sort > out_mkdup_sorted.bam.
```

FrickTobias:
The reads which are not marked as read duplicates, do they seem to be duplicates? I'd say whether they are read duplicates or not is the most important factor for deciding which we should go for.

pontushojer:
>The reads which are not marked as read duplicates, do they seem to be duplicates? I'd say whether they are read duplicates or not is the most important factor for deciding which we should go for.

True, I am investigating this further.

pontushojer:
From [supplementary](https://www.dropbox.com/s/s7vvdtf2gmytvay/SAMBLASTER_Supplemental.pdf?dl=0)

[...]

PR #132, opened 2019-10-29
Implement alternative duplicate markers

pontushojer:
Here is the implementation of sambamba and samblaster as duplicate markers. Picard MarkDuplicates has been removed as it was proven to be quite resource intensive and slow (see #113).

Note that samblaster is integrated as pipe in the mapping step as this is the intended benefit of the tool. Thus it will skip the `mark_duplicates` step and instead go from `tagbam` directly to `clusterrmdup`. The naming is a bit off as the file `mapped.sorted.bam` now has duplicates marked. I think this is fine for now and we can change it whenever we select a specific tool for marking duplicates. But comment if you have any objections or suggestions!

### DIFF
--- a/src/blr/Snakefile
+++ b/src/blr/Snakefile
@@ -95,21 +95,17 @@ rule map_reads:
                 " {input.r2_fastq}",
         }
         command = commands[config["read_mapper"]].format(**locals(), **globals())
-        if config["duplicate_marker"] == "samblaster":
-            shell(
-                "{command} 2> >(tee {log} >&2) |"
-                "samblaster |"
-                "samtools sort -"
-                " -@ {threads}"
-                " -O BAM > {output.bam}"
-            )
-        else:
-            shell(
-                "{command} 2> >(tee {log} >&2) |"
-                "samtools sort -"
-                " -@ {threads}"
-                " -O BAM > {output.bam}"
-            )
+
+        # Add duplicate marking to map_reads pipe if samblaster selected.
+        mkdup = "samblaster |" if config["duplicate_marker"] == "samblaster" else ""
+
+        shell(
+            "{command} 2> >(tee {log} >&2) |"
+            "{mkdup}"
+            "samtools sort -"
+            " -@ {threads}"
+            " -O BAM > {output.bam}"
+        )
 
 
 rule tagbam:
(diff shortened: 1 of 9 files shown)


src/blr/Snakefile
```diff
@@ -96,18 +95,27 @@ rule map_reads:
...
         }
         command = commands[config["read_mapper"]].format(**locals(), **globals())
-        shell(
-            "{command} 2> >(tee {log} >&2) |"
-            "samtools sort -"
-            " -@ {threads}"
-            " -O BAM > {output.bam}"
-        )
+        if config["duplicate_marker"] == "samblaster":
+            shell(
```
marcelm: You can avoid the duplication of the entire shell command by setting a variable to `"samblaster |"` if the configured duplicate marker is samblaster and to `""` otherwise. Then just use that variable in the string that is passed to the shell command.

pontushojer: That is true 👍

src/blr/blr.yaml
```diff
@@ -4,7 +4,8 @@ molecule_tag: MI  # Used to store molecule ID, same as 10x default.
 num_mol_tag: MN # Used to store number of molecules per barcode
 sequence_tag: RX    # Used to store original barcode sequence in bam file. 'RX' is 10x genomic default
 genome_reference:  # Path to indexed reference
-read_mapper: bowtie2  # Choose bwa, bowtie2 or minimap2
+read_mapper: bowtie2  # Choose bwa or bowtie2 or minimap2
```
marcelm: Hm, why did you change the grammar here? This was quite standard English before.

pontushojer: I think I made a mistake while rebasing, will change.

src/blr/Snakefile
```diff
@@ -123,15 +131,13 @@ rule mark_duplicates:
...
-        stderr = "picard_mkdup.log"
+    log: "mark_duplicates.log"
+    threads: 20
     shell:
-        "picard -Xms{config[heap_space]}g MarkDuplicates"
-        " I={input.bam}"
-        " O={output.bam}"
-        " M={log.metrics}"
-        " ASSUME_SORT_ORDER=coordinate 2> {log.stderr}"
+        "sambamba markdup "
```
marcelm: Please remove the space before the last `"` in this and the next line.

pontushojer: Fixed!

FrickTobias:
Loos great!

Issue #133, opened 2019-10-30
tagfastq: cannot allocate memory


FrickTobias:
# Problem

I'm trying to run library with 440 M read pairs and I'm getting a traceback due to being unable to allocate memory in the `tagfastq` step (see log file at the bottom).

Since Cutadapt works fine with the given amount of reads I'm a bit confused as to what might be giving this error, are we misusing `dnaio` or could this be because we've already allocated a lot of memory to holding the barcoding information?

If I remember correctly `cutadapt` threads by assigning a specific core for reading, could this be related since the traceback specifies it **cannot allocate** memory (rather than running out of it).

I'm running the pipe with 20 cores, 128 Gb RAM.

### Barcode reading functions
```
def parse_corrected_barcodes(open_file):
    """
    Parse starcode cluster output and return a dictionary with raw sequences pointing to a
    corrected canonical sequence
    :param open_file: starcode tabular output file.
    :return: dict: raw sequences pointing to a corrected canonical sequence.
    """
    corrected_barcodes = dict()
    for cluster in tqdm(open_file, desc="Clusters processed"):
        canonical_seq, _, cluster_seqs = cluster.strip().split("\t", maxsplit=3)
        corrected_barcodes.update({raw_seq: canonical_seq for raw_seq in cluster_seqs.split(",")})
    return corrected_barcodes
```
```
def parse_raw_barcodes(open_file):
    """
    Parse FASTA/FASTQ containing barcode sequences and return a dictionary with entry headers
    pointing to a raw barcodes sequence.
    :param open_file: dnaio odject.
    :return: dict: entry headers pointing to a raw barcodes sequence
    """
    raw_barcodes = dict()
    for barcode in tqdm(open_file, desc="Raw barcodes processed"):
        header, _ = barcode.name.split(maxsplit=1)
        raw_barcodes[header] = barcode.sequence
    return raw_barcodes
```
### Loop at traceback

```
 with dnaio.open(args.input1, file2=args.input2, interleaved=in_interleaved, mode="r") as reader, \
            dnaio.open(args.output1, file2=args.output2, interleaved=out_interleaved, mode="w") as writer:
        for read1, read2 in tqdm(reader, desc="Read pairs processed"):
            # Header parsing
            name_and_pos_r1, read_and_index_r1 = read1.name.split(maxsplit=1)
            name_and_pos_r2, read_and_index_r2 = read2.name.split(maxsplit=1)

            try:
[...]

PR #134, opened 2019-10-31
tagfastq rewrite to improve performance. 

pontushojer:
Raw barcodes are not read along with the reads FASTQs as they should sorted in the same order. Made a search function that can adapt to missing barcodes and headers, currently it buffers ahead up to 10 entries which should be enough as long as we don't have more then 10 missing read entries in succession.

See issue #133

### DIFF
--- a/src/blr/cli/tagfastq.py
+++ b/src/blr/cli/tagfastq.py
@@ -50,7 +50,8 @@ def main(args):
     reads_missing_barcode = 0
     separator = args.sep
     # Parse input FASTA/FASTQ for read1 and read2, raw barcodes and write output
-    with dnaio.open(args.input1, file2=args.input2, interleaved=in_interleaved, mode="r", fileformat="fastq") as reader, \
+    with dnaio.open(args.input1, file2=args.input2, interleaved=in_interleaved, mode="r",
+                    fileformat="fastq") as reader, \
             dnaio.open(args.output1, file2=args.output2, interleaved=out_interleaved, mode="w",
                        fileformat="fastq") as writer, \
             dnaio.open(args.raw_barcodes, mode="r") as raw_bc_reader:
--- a/src/blr/cli/tagfastq.py
+++ b/src/blr/cli/tagfastq.py
@@ -51,7 +51,8 @@ def main(args):
     separator = args.sep
     # Parse input FASTA/FASTQ for read1 and read2, raw barcodes and write output
     with dnaio.open(args.input1, file2=args.input2, interleaved=in_interleaved, mode="r", fileformat="fastq") as reader, \
-            dnaio.open(args.output1, file2=args.output2, interleaved=out_interleaved, mode="w", fileformat="fastq") as writer, \
+            dnaio.open(args.output1, file2=args.output2, interleaved=out_interleaved, mode="w",
+                       fileformat="fastq") as writer, \
             dnaio.open(args.raw_barcodes, mode="r") as raw_bc_reader:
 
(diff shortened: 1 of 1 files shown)


FrickTobias:
Looks nice, I think this should work as a solution. I'll post the a pre/post memory usage graph when I've tested it.

**My minor edit:** I edited the row length down so it follows our flake8/pep8 rules (and I did it twice since my editor was set to a higher upper limit than flake8).

src/blr/cli/tagfastq.py
```diff
@@ -88,6 +89,22 @@ def main(args):
     logger.info("Finished")
 
 
+def search_bc(iterator: iter, header: str, cache: dict, maxiter: int = 10):
```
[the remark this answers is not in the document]
marcelm: You could rename `iterator` to `uncorrected_barcodes` to make it clearer what this represents.

src/blr/cli/tagfastq.py
```diff
@@ -88,6 +89,22 @@ def main(args):
...
+    # Check it header is stored in cache. If not move forward on step in iterator at look again.
+    iteration = 0
+    while header not in cache and iteration < maxiter:
```
marcelm: Note that every uncorrected barcode that exists in the input barcodes, but is never searched for, will remain in the cache. Perhaps this doesn’t matter that much, but you could check whether the value you get from `next(iterator)` is the one you are searching for and then clear the cache because you know that you’ll not need the earlier barcodes anymore.

src/blr/cli/tagfastq.py
```diff
@@ -88,6 +89,22 @@ def main(args):
...
+        iteration += 1
+        # Progress iterator
+        try:
+            cache.update(next(iterator))
+        except StopIteration:
+            break
+
+    barcode_seq = cache.pop(header, None)
```
marcelm: Remove either the empty line before or after this line.

src/blr/cli/tagfastq.py
```diff
@@ -109,11 +126,9 @@ def parse_raw_barcodes(open_file):
     :param open_file: dnaio odject.
     :return: dict: entry headers pointing to a raw barcodes sequence
```
marcelm: The docstring has become out-of-sync with what the function actually returns.

src/blr/cli/tagfastq.py
```diff
@@ -88,6 +89,22 @@ def main(args):
     logger.info("Finished")
 
 
+def search_bc(iterator: iter, header: str, cache: dict, maxiter: int = 10):
+    # Check it header is stored in cache. If not move forward on step in iterator at look again.
+    iteration = 0
+    while header not in cache and iteration < maxiter:
+        iteration += 1
+        # Progress iterator
```
marcelm: Initially I read this as having something to do with a progress report ... I think "advance iterator" would be more customary. And consider whether the comment is necessary at all. The `next(iterator)` already describes exactly that.

src/blr/cli/tagfastq.py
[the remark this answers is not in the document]
pontushojer: Good that you pointed that out! I have not read up too much on these, the mypy resource is great. In that case I will remove the `iter`and `dict` annotations.

src/blr/cli/tagfastq.py
pontushojer: Implemented in new PR.

pontushojer:
@marcelm This was merged before I was able to add the requested changes. I am opening a new PR to include them.

PR #135, opened 2019-11-05
Use HapCUT2 from Bioconda

### LINKED ISSUE #122: Simplify how binaries are found

marcelm:
I have written a HapCUT2 recipe for Bioconda, which was merged yesterday. This PR makes use of the Bioconda package. It also removes the `install-hapcut2.sh` script and the `hapcut2` configuration
setting.

Closes #122.


FrickTobias:
Awesome!

PR #136, opened 2019-11-06
Tagfastq improvement - additional edits

pontushojer:
This includes edits from review on PR #134 that was missed before the merge. 

Edits include:
- Style changes
- Naming
- Documentation

I also improved the readability of the `search_barcodes` function. This also includes the addition that the cache is dropped is the reads and barcodes files are synced. Previously all barcodes not searched for in the cache (i.e. missing in the reads) were kept. They were not many, 38 in my 1 million test file but for larger files this would soak up some additional memory.

### DIFF
--- a/src/blr/cli/tagfastq.py
+++ b/src/blr/cli/tagfastq.py
@@ -122,7 +122,7 @@ class BarcodeReader:
                 self._cache.clear()
                 return barcode_sequence
 
-            self._cache.update({barcode_read_name: barcode_sequence})
+            self._cache[barcode_read_name] = barcode_sequence
         return None
 
     def __enter__(self):
--- a/src/blr/cli/tagfastq.py
+++ b/src/blr/cli/tagfastq.py
@@ -46,31 +46,32 @@ def main(args):
     out_interleaved = not args.output2
     logger.info(f"Output detected as {'interleaved' if out_interleaved else 'paired'} FASTQ.")
 
-    raw_barcodes_cache = dict()
+    uncorrected_barcodes_cache = dict()
     reads_missing_barcode = 0
     separator = args.sep
-    # Parse input FASTA/FASTQ for read1 and read2, raw barcodes and write output
+    # Parse input FASTA/FASTQ for read1 and read2, uncorrected barcodes and write output
     with dnaio.open(args.input1, file2=args.input2, interleaved=in_interleaved, mode="r",
                     fileformat="fastq") as reader, \
             dnaio.open(args.output1, file2=args.output2, interleaved=out_interleaved, mode="w",
                        fileformat="fastq") as writer, \
-            dnaio.open(args.raw_barcodes, mode="r") as raw_bc_reader:
+            dnaio.open(args.uncorrected_barcodes, mode="r") as uncorrected_barcode_reader:
 
-        raw_bc_iterator = parse_raw_barcodes(raw_bc_reader)
+        uncorrected_barcode_iterator = parse_uncorrected_barcodes(uncorrected_barcode_reader)
 
         for read1, read2 in tqdm(reader, desc="Read pairs processed"):
             # Header parsing
             name_and_pos_r1, read_and_index_r1 = read1.name.split(maxsplit=1)
             name_and_pos_r2, read_and_index_r2 = read2.name.split(maxsplit=1)
 
-            raw_barcode_seq, raw_barcodes_cache = search_bc(raw_bc_iterator, name_and_pos_r1, raw_barcodes_cache)
+            uncorrected_barcode_seq = search_barcode(uncorrected_barcode_iterator, name_and_pos_r1,
+                                                     uncorrected_barcodes_cache)
 
             # Check if barcode was found and update header with barcode info.
-            if raw_barcode_seq:
-                corr_barcode_seq = corrected_barcodes[raw_barcode_seq]
+            if uncorrected_barcode_seq:
+                corrected_barcode_seq = corrected_barcodes[uncorrected_barcode_seq]
 
-                raw_barcode_id = f"{args.sequence_tag}:Z:{raw_barcode_seq}"
-                corr_barcode_id = f"{args.barcode_tag}:Z:{corr_barcode_seq}"
+                raw_barcode_id = f"{args.sequence_tag}:Z:{uncorrected_barcode_seq}"
+                corr_barcode_id = f"{args.barcode_tag}:Z:{corrected_barcode_seq}"
 
                 # Create new name with barcode information.
                 new_name = separator.join([name_and_pos_r1, raw_barcode_id, corr_barcode_id])
@@ -84,25 +85,24 @@ def main(args):
             # Write to out
             writer.write(read1, read2)
 
+    logger.info(f"Barcodes missing trimmed reads: {len(uncorrected_barcodes_cache)}")
     logger.info(f"Read-pairs missing barcodes: {reads_missing_barcode}")
 
     logger.info("Finished")
 
 
-def search_bc(iterator: iter, header: str, cache: dict, maxiter: int = 10):
-    # Check it header is stored in cache. If not move forward on step in iterator at look again.
+def search_barcode(uncorrected_barcodes, header: str, cache, maxiter: int = 10):
+    # Check it header is stored in cache. If not move forward on step in uncorrected_barcodes at look again.
     iteration = 0
     while header not in cache and iteration < maxiter:
         iteration += 1
-        # Progress iterator
         try:
-            cache.update(next(iterator))
+            cache.update(next(uncorrected_barcodes))
         except StopIteration:
             break
 
     barcode_seq = cache.pop(header, None)
-
-    return barcode_seq, cache
+    return barcode_seq
 
 
 def parse_corrected_barcodes(open_file):
@@ -119,12 +119,12 @@ def parse_corrected_barcodes(open_file):
     return corrected_barcodes
 
 
-def parse_raw_barcodes(open_file):
+def parse_uncorrected_barcodes(open_file):
     """
     Parse FASTA/FASTQ containing barcode sequences and return a dictionary with entry headers
     pointing to a raw barcodes sequence.
     :param open_file: dnaio odject.
-    :return: dict: entry headers pointing to a raw barcodes sequence
+    :return: iterable with dict elements
     """
     for barcode in tqdm(open_file, desc="Raw barcodes processed"):
         header, _ = barcode.name.split(maxsplit=1)
@@ -133,8 +133,8 @@ def parse_raw_barcodes(open_file):
 
 def add_arguments(parser):
     parser.add_argument(
-        "raw_barcodes",
-        help="FASTQ/FASTA for raw barcodes.")
+        "uncorrected_barcodes",
+        help="FASTQ/FASTA for uncorrected barcodes.")
     parser.add_argument(
         "corrected_barcodes",
         help="FASTQ/FASTA for error corrected barcodes. Currently accepts output from starcode "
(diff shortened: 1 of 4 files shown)


src/blr/cli/tagfastq.py
```diff
@@ -89,20 +90,21 @@ def main(args):
...
-        # Progress iterator
-        try:
-            cache.update(next(iterator))
-        except StopIteration:
-            break
+def search_barcode(uncorrected_barcodes, header: str, cache, maxiter: int = 10):
+    if header in cache:
+        return cache.pop(header)
+
+    for iteration, header_barcode_pair in enumerate(uncorrected_barcodes):
```
marcelm: You could avoid the extra `iteration` variable by using `islice` here (`from itertools import islice`):
```
for header_barcode_pair in islice(uncorrected_barcodes, maxiter):
```
Then also the `if iteration >= maxiter:` below isn’t needed.

src/blr/cli/tagfastq.py
```diff
@@ -46,31 +46,32 @@ def main(args):
...
-        raw_bc_iterator = parse_raw_barcodes(raw_bc_reader)
+        uncorrected_barcode_iterator = parse_uncorrected_barcodes(uncorrected_barcode_reader)
 
         for read1, read2 in tqdm(reader, desc="Read pairs processed"):
             # Header parsing
             name_and_pos_r1, read_and_index_r1 = read1.name.split(maxsplit=1)
             name_and_pos_r2, read_and_index_r2 = read2.name.split(maxsplit=1)
 
-            raw_barcode_seq, raw_barcodes_cache = search_bc(raw_bc_iterator, name_and_pos_r1, raw_barcodes_cache)
+            uncorrected_barcode_seq = search_barcode(uncorrected_barcode_iterator, name_and_pos_r1,
```
marcelm: Have you considered factoring out a class? This is totally optional and I just want to mention it since I noticed it now, but if you create a class `BarcodeReader`/`BarcodeCache` or similar, then you could move the cache dict, the iterator, the barcode parser and the `search_barcode` function into it. This would nicely encapsulate the barcode reading functionality. From the main function, you would only need to call the constructor and then call `barcode_cache.retrieve(name_and_pos_r1)` or similar.

src/blr/cli/tagfastq.py
```diff
@@ -119,12 +121,12 @@ def parse_corrected_barcodes(open_file):
...
-def parse_raw_barcodes(open_file):
+def parse_uncorrected_barcodes(open_file):
     """
     Parse FASTA/FASTQ containing barcode sequences and return a dictionary with entry headers
     pointing to a raw barcodes sequence.
     :param open_file: dnaio odject.
-    :return: dict: entry headers pointing to a raw barcodes sequence
+    :return: iterable with dict elements
     """
     for barcode in tqdm(open_file, desc="Raw barcodes processed"):
```
marcelm: Doesn’t this `tqdm` now interfere with the other progress report in line 61?

src/blr/cli/tagfastq.py
pontushojer: Sweet! I will add that.

src/blr/cli/tagfastq.py
pontushojer: Not really, check this output from the latest run:

```
tagfastq - INFO: Starting
Clusters processed: 7774460it [00:21, 356120.73it/s]
tagfastq - INFO: Input detected as paired FASTQ.
tagfastq - INFO: Output detected as paired FASTQ.
Read pairs processed: 437146385it [2:11:42, 55314.71it/s]
tagfastq - INFO: Read-pairs missing barcodes: 4030880      
tagfastq - INFO: Finished
Raw barcodes processed: 433168106it [2:11:44, 54801.86it/s]
```

The buffer from "Read pairs processed" and "Raw barcodes processed" appear next to each other on separate lines during the run. After the run however the "Raw barcodes processed" is move below the other outputs, possibly as the iterator is not formally exhausted.

marcelm: Ok, I guess tqdm is smart about this. Good!

src/blr/cli/tagfastq.py
pontushojer: I did an attempt on this. I have not work much with classes before, so if I am doing something bad or against conventions please inform me :)

marcelm: Looks quite good already, I’ll just add a couple of comments

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,42 @@ def parse_corrected_barcodes(open_file):
...
-    :return: dict: entry headers pointing to a raw barcodes sequence
-    """
-    for barcode in tqdm(open_file, desc="Raw barcodes processed"):
-        header, _ = barcode.name.split(maxsplit=1)
-        yield {header: barcode.sequence}
+class BarcodeReader:
+    def __init__(self, filename):
+        self.cache = dict()
+        self.filename = filename
+        self.barcodes = self.parse()
```
marcelm: To mark these attributes as "private", you should prefix them with an underscore (`_cache`, etc.).

marcelm: It’s a bit unusual that the constructor leaves the object in a half-initialized state (`self.open_file` is set in `__enter__`). You should open the input file already here. Then, if you add a `close()` method as well, it is possible to close the `BarcodeReader` without using it as a context manager.

marcelm: And I think it’s enough to call the file `_file`. The expectation would be anyway that it is open.

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,42 @@ def parse_corrected_barcodes(open_file):
...
+
+            self.cache.update(header_barcode_pair)
+        return None
+
+    def __enter__(self):
+        self.open_file = dnaio.open(self.filename, mode="r")
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return self.open_file.close()
```
marcelm: As mentioned above, it would be nicer to implement a separate `close()` method. Then you can defer to that for closing. So this could then be replaced with just `self.close()`.
Since `close()` doesn’t return anything, you should also omit the `return`.

src/blr/cli/tagfastq.py
pontushojer: Thanks for the comments! I implemented the changes in the new commit.

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,44 @@ def parse_corrected_barcodes(open_file):
...
+            return self._cache.pop(header)
+
+        for header_barcode_pair in islice(self.barcodes, maxiter):
+            # If header in next pair then parser lines are synced --> drop cache.
+            if header in header_barcode_pair:
+                self._cache.clear()
+                return header_barcode_pair[header]
+
+            self._cache.update(header_barcode_pair)
+        return None
```
FrickTobias: I think this could be written with less return statements. I think it's easier to have an overview if there is only one return statement.

```suggestion
    def get_barcode(self, header, maxiter=10):
        if header in self._cache:
            barcode = self._cache.pop(header)
        else:
            for header_barcode_pair in islice(self.barcodes, maxiter):
                # If header in next pair then parser lines are synced --> drop cache.
                if header in header_barcode_pair:
                    self._cache.clear()
                    barcode = header_barcode_pair[header]
                self._cache.update(header_barcode_pair)
        return barcode
```

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,44 @@ def parse_corrected_barcodes(open_file):
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
```
FrickTobias: I guess this is no longer needed right?

src/blr/cli/tagfastq.py
pontushojer: I actually prefer mine with the manny return statements. This avoids assigning a value to the `barcode` variable before returning it.  Mine is also less complex in my opinion with less indentation levels. 

Note also add that in order to  have the same functionality as my code the rewrite would have to look a bit more like this:

```python
def get_barcode(self, header, maxiter=10):
   if header in self._cache:
      barcode = self._cache.pop(header)
   else:
      barcode = None
      for header_barcode_pair in islice(self.barcodes, maxiter):
         # If header in next pair then parser lines are synced --> drop cache.
         if header in header_barcode_pair:
            self._cache.clear()
            barcode = header_barcode_pair[header]
            break
         self._cache.update(header_barcode_pair)
   return barcode
```
The `barcode = None` needs to be included in the case that the barcode is not present. The `break` should be included if you want to stop parsing if the barcode is found. If not included the cache will be filled 10 barcodes at the time including any matched barcode. 

I tested running 10 million reads using your code without the `break` and it works fine. Performance similar to my code.

src/blr/cli/tagfastq.py
pontushojer: This is still needed in order for it to function as a context manager.  Context managers require both `__enter__` and `__exit__` functions (see [PEP 343](https://www.python.org/dev/peps/pep-0343/)).

marcelm:
Hi, can you please wait with merging? I have a couple of additional comments, but am already home now.

src/blr/cli/tagfastq.py
marcelm: I also think @pontushojer’s version is fine, but for completeness, initializing `barcode` to `None` could be done by using a default value for `pop`:
```
   barcode = self._cache.pop(header, None)
   if barcode is None:
```

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,44 @@ def parse_corrected_barcodes(open_file):
...
+class BarcodeReader:
+    def __init__(self, filename):
+        self._cache = dict()
+        self._file = dnaio.open(filename, mode="r")
+        self.barcodes = self.parse()
+
+    def parse(self):
+        for barcode in tqdm(self._file, desc="Uncorrected barcodes processed"):
+            header, _ = barcode.name.split(maxsplit=1)
+            yield {header: barcode.sequence}
```
marcelm: I suggest to call it `read_name` or perhaps `read_id` instead of `header` ("header" would imply for me that it includes the entire line, even including the `@`).

marcelm: I also suggest that you return a pair `(read_name, barcode.sequence)` here.

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,44 @@ def parse_corrected_barcodes(open_file):
...
+
+    def get_barcode(self, header, maxiter=10):
+        if header in self._cache:
+            return self._cache.pop(header)
+
+        for header_barcode_pair in islice(self.barcodes, maxiter):
```
marcelm: This is currently not actually a pair, but a dict. If you actually let `parse()` return pairs, then you can write 
```
for read_name, barcode in islice(...):`
```

src/blr/cli/tagfastq.py
FrickTobias: Alright

FrickTobias:
Great feedback, do you think it is ready to be merged? @marcelm

src/blr/cli/tagfastq.py
```diff
@@ -119,22 +101,44 @@ def parse_corrected_barcodes(open_file):
...
+        if read_name in self._cache:
+            return self._cache.pop(read_name)
+
+        for barcode_read_name, barcode_sequence in islice(self.barcodes, maxiter):
+            # If read_name in next pair then parser lines are synced --> drop cache.
+            if read_name == barcode_read_name:
+                self._cache.clear()
+                return barcode_sequence
+
+            self._cache.update({barcode_read_name: barcode_sequence})
```
marcelm: This should simply be `self._cache[barcode_name] = barcode_sequence`

marcelm:
I noticed one more tiny thing, but if that is taken care of, this is good to go!

src/blr/cli/tagfastq.py
pontushojer: Fixed!

Issue #116, opened 2019-10-07
Evaluate where more streaming can be used


marcelm:
More streaming can perhaps be employed. Currently, some large intermediate files are generated, some uncompressed, some compressed. This is slow: When writing them uncompressed, file I/O performance is likely a bottleneck. When writing them compressed, that takes a lot of CPU hours. Instead, one can let one program write uncompressed and send the output to the next program via a pipe. This avoids the intermediate file to disk entirely.

Perhaps this can be used for some of the intermediate BAM files. (And the other non-final ones should be marked as `temp()` as suggested in #115).

pontushojer:
As noted in #133 we can pipe directly from the read trimming into the `tagfastq` script thereby skipping the intermediate `trimmed.1.fastq.gz` and `trimmed.2.fastq.gz` files.

pontushojer:
I looked into the four steps `tagbam`, `clusterrmdup`, `buildmolecules` and `filterclusters` and this in our current possibilities for streaming: 
- **`tagbam`: Streaming possible**
- `clusterrmdup`: No streaming
- `buildmolecules`: No streaming
- **`filterclusters`: Streaming possible**

`clusterrmdup` and `buildmolecules` read in the file two times and thus require the input to be read from disc.

PR #137, opened 2019-11-06
Implement streaming of trimmed fastq into `tagfastq`

pontushojer:
Removes intermediate `trimmed.1.fastq.gz` and `trimmed.2.fastq.gz` files.

Related to issue #116

### DIFF
--- a/doc/develop.rst
+++ b/doc/develop.rst
@@ -35,4 +35,4 @@ where that information is defined.
     "MI", "-m, --molecule-tag", "SAM tag for storing molecule index specifying a identified molecule
     for each barcode."
     "MN", "-n, --number-tag", "SAM tag for storing molecule count for a particular barcode."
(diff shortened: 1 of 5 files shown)


FrickTobias:
What do you think about naming the files like `trimmed.barcoded` rather than `trimmed_barcoded` in order to be consistent with the current naming convention used for the BAM files?

pontushojer:
> What do you think about naming the files like `trimmed.barcoded` rather than `trimmed_barcoded` in order to be consistent with the current naming convention used for the BAM files?

Sure! That is more consistent

pontushojer:
I thought this was PR #136 so I added the name change to that PR by mistake. But I can rebase this one after the other PR is merged. So **hold of on merging this for now!!**

FrickTobias:
PR #136 is now merged. 

By the way, I think you can rebase it to the branch before merging since git should the first PRs commit hashes as already merged.

Issue #77, opened 2019-09-09
Common fetch_bc function


pontushojer:
Issue from discussion in #76

> One thing that I thought about was maybe to move the `fetch_tag` function used into >`utils`. The function is used quite similar in both `filterclusters` and `buildmolecules` >and could possibly be useful for other programs. The differ abit between them at the >moment but I think you could combine them like this.

>```python
>def fetch_bc(pysam_read, tag, summary=None, default=None):
>    """
>    Fetches barcode from a bam file tag, returns None if reads isn't tagged.
>    """
>
>    try:
>        result = pysam_read.get_tag(tag)
>    except KeyError:
>        result = default
>
>        if summary:
>            summary.reads_without_tag += 1
>
>    return result
>```

pontushojer:
PR #141 solves

PR #141, opened 2019-11-12
Common fetch bam tag function

FrickTobias:
Solves #77 

Common fetch bam tag function which returns tag value or None.


src/blr/cli/buildmolecules.py
```diff
@@ -74,7 +76,7 @@ def build_molecules(pysam_openfile, barcode_tag, window, min_reads, summary):
             continue
 
         # Fetches barcode and genomic position. Position will be formatted so start < stop.
-        barcode = fetch_bc(pysam_read=read, barcode_tag=barcode_tag, summary=summary)
+        barcode = utils.get_bamtag(pysam_read=read, tag=barcode_tag)
         if barcode and not read.is_unmapped:
             read_start, read_stop = sorted((read.reference_start, read.reference_end))
```
pontushojer:
```suggestion
elif not barcode:
   summary.reads_without_barcode += 1
```

Since this was removed in the local function we can add this to continue to track reads without barcodes.

FrickTobias: I also moved the `read.is_unmapped` check to the top so the main loop could be put in the `else` clause instead.

src/blr/utils.py
```diff
@@ -38,3 +38,17 @@ def guess_paired_path(path: Path):
...
+    Fetches tags from bam files. Return an empty value of the same type if not found.
+    :param pysam_read: pysam read object
+    :param tag: bam tag to fetch
+    :param type: what type to return if read does not have the tag
+    :return: bam tag value
+    """
+    try:
+        return pysam_read.get_tag(tag)
+    except KeyError:
+        return None
```
pontushojer: Now `None` is returned by default by the `utils.get_tag` function upon `KeyError`. Do you think it would be useful to be able to specify a default?

FrickTobias: I used that from the beginning but found it better to return `None`. 

In cases where we have want to fetch a `tag_value` but there it is not defined, the original need of the `tag_val` should reasonably mean an absence of it should be handled differently.

PR #143, opened 2019-11-14
Changes to help facilitate streaming

pontushojer:
Related to issue #116 

 - Enable writing to stdout in SAM or BAM format for `tagbam`, `clusterrmdup`, `buildmolecules` and `filterclusters`.
 - Updated information in docstrings and help messages. Included info on which scripts can handle reading from stdin (currently `tagbam` and `filterclusters`.)

### DIFF
--- a/src/blr/cli/buildmolecules.py
+++ b/src/blr/cli/buildmolecules.py
@@ -17,7 +17,6 @@ logger = logging.getLogger(__name__)
 
 def main(args):
     summary = Summary()
-    out_mode = utils.get_output_mode(args.output, args.output_format)
 
     # Build molecules from BCs and reads
     with pysam.AlignmentFile(args.input, "rb") as infile:
@@ -28,7 +27,7 @@ def main(args):
 
     # Writes filtered out
     with pysam.AlignmentFile(args.input, "rb") as openin, \
-            pysam.AlignmentFile(args.output, out_mode, template=openin) as openout:
+            pysam.AlignmentFile(args.output, "wb", template=openin) as openout:
         logger.info("Writing filtered bam file")
         for read in tqdm(openin.fetch(until_eof=True)):
             name = read.query_name
@@ -276,10 +275,7 @@ def add_arguments(parser):
                              "-b/--barcode-tag.")
 
     parser.add_argument("-o", "--output", default="-",
-                        help="Write output SAM/BAM to file rather then stdout. Format is inferred from the "
-                             "filename unless -O/--output-format is specified.")
-    parser.add_argument("-O", "--output-format", choices=["SAM", "BAM"],
-                        help="Specify output format.")
+                        help="Write output BAM to file rather then stdout.")
     parser.add_argument("-t", "--threshold", type=int, default=4,
                         help="Threshold for how many reads are required for including given molecule in statistics "
                              "(except_reads_per_molecule). Default: %(default)s")
--- a/src/blr/cli/buildmolecules.py
+++ b/src/blr/cli/buildmolecules.py
@@ -1,5 +1,5 @@
 """
-Tags .bam file with molecule information based on barcode sequence and genomic proximity.
+Tags SAM/BAM file with molecule information based on barcode sequence and genomic proximity.
 
 A molecule is defined by having 1) minimum --threshold reads and including all reads with the same barcode which are 2)
 a maximum distance of --window between any given reads.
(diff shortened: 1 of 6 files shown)


src/blr/utils.py
```diff
@@ -52,3 +52,21 @@ def get_bamtag(pysam_read, tag):
...
+    :param spec_format: Should be either SAM or BAM.
+    :return: String containing the pySAM mode information. String is 'w' for SAM and 'wb' for BAM.
+    """
+    format_to_mode = {
+        "SAM": "w",
+        "BAM": "wb"
+    }
+
+    if filename != "-":
+        spec_format = filename.split(".")[-1].upper()
```
FrickTobias: This seem to work on filenames rather than `spec_format` names so I think it might be more clear if it says `filename` or similar instead since that is what it assumes.

src/blr/cli/tagbam.py
```diff
@@ -47,10 +48,16 @@ def main(args):
...
-                        help="BAM file with mapped reads which is to be tagged with barcode ids.")
-    parser.add_argument("output_tagged_bam",
-                        help="BAM file with barcode cluster id in the bc tag.")
+    parser.add_argument("input",
+                        help="SAM/BAM file with mapped reads which is to be tagged with barcode information. To read "
+                             "from stdin use '-'.")
+
+    parser.add_argument("-o", "--output", default="-",
+                        help="SAM/BAM file with barcodes information in the specified SAM tags. Default: write to "
+                             "stdout.")
```
FrickTobias: Since we already have the `get_output_mode` in `utils` which can discern file format from the name I think it's better if we instead assume output format from filenames.

FrickTobias:
Hm, I think I need to read this again, I misunderstood some stuff.

src/blr/cli/buildmolecules.py
```diff
@@ -270,12 +271,17 @@ def write_molecule_stats(self, output_prefix, molecule_dict):
...
-
+    parser.add_argument("input",
+                        help="Sorted SAM/BAM file tagged with barcode in the same tag as specified in "
+                             "-b/--barcode-tag.")
+
+    parser.add_argument("-o", "--output", default="-",
+                        help="Output SAM/BAM file with molecule tags found under the tag specified at "
+                             "-m/--molecule-tag and molecules number for each barcode under the specified "
+                             "-n/--number-tag. Default: write to stdout.")
+    parser.add_argument("-O", "--output-format", choices=["SAM", "BAM"], default="SAM",
```
FrickTobias: To me it makes more sense to infer name from output file and if `-O` is specified that is overridden.

src/blr/cli/buildmolecules.py
```diff
@@ -17,17 +17,18 @@
 
 def main(args):
     summary = Summary()
+    out_mode = utils.get_output_mode(args.output, args.output_format)
```
FrickTobias: If we have an option for specifying output format I think that should override the filename.

```suggestion
    if not args.output_format:
        out_mode = utils.get_output_mode(args.output, args.output_format)
```

src/blr/utils.py
```diff
@@ -52,3 +52,21 @@ def get_bamtag(pysam_read, tag):
...
+    return format_to_mode[spec_format]
```
FrickTobias: Then we can remove the `if` syntax here since it only get the output mode if the user is not specifying it in an option.

pontushojer:
I have now changed the logic for how the `-o/--output` and `-O/--output-format` interacts. The specified format now overrides the file name. Also if no format or file name is given output is written to stdout as SAM format.

marcelm:
I don’t have time to look into this in full detail, but have a comment anyway.

I think this is getting too complicated. Why have the option of selecting the output format when you’re using just one anyway in the pipeline? These scripts don’t need to be universally usable. If you make them so, you’ll also have to test that all the options actually work!

Just choose a consistent way in which they work: Always output SAM on standard output, for example. This is consistent with how many read mappers work. Alternatively, using uncompressed BAM is even better for efficient streaming.

The complications of having to handle different file formats and different modes (standard output/input vs. reading/writing to disk) need to be handled at the user-facing "edges" of the pipeline where we don’t have control over the input or output. Internally, we can pick one representation and stick to it.

pontushojer:
> I don’t have time to look into this in full detail, but have a comment anyway.
> 
> I think this is getting too complicated. Why have the option of selecting the output format when you’re using just one anyway in the pipeline? These scripts don’t need to be universally usable. If you make them so, you’ll also have to test that all the options actually work!
> 
> Just choose a consistent way in which they work: Always output SAM on standard output, for example. This is consistent with how many read mappers work. Alternatively, using uncompressed BAM is even better for efficient streaming.
> 
> The complications of having to handle different file formats and different modes (standard output/input vs. reading/writing to disk) need to be handled at the user-facing "edges" of the pipeline where we don’t have control over the input or output. Internally, we can pick one representation and stick to it.

My thinking was that the different options would provide flexibility if we wanted to stream output in the future. But this may in fact have made thing more complicated than is required. I have looked as the scripts more as "tools" as a part of the package rather then processing step in a defined pipeline. For our purposes one defined input/output format is sufficient. This I will change.

marcelm:
> I have looked as the scripts more as "tools" as a part of the package rather then processing step in a defined pipeline.

Yes, from that point of view, these changes are understandable. However, look up the YAGNI principle ;-). (You Ain’t Gonna Need It.) More code that you don’t need at the moment is just a maintenance burden.

pontushojer:
That is definitely a principle I need to consider more often! :)

PR #145, opened 2019-11-19
add freebayes to env

FrickTobias:
`freebayes` was not included in the environment file before so I added it.


FrickTobias:
I'm considering this a minor change so I'll go ahead and merge (with only one approval).

marcelm:
Looks good to me, sorry for late reply. (But I think these almost trivial things are fine to merge without everyone looking over them.)

Issue #117, opened 2019-10-07
Test whether Picard SamToFastq should write gzip directly


marcelm:
SamToFastq can write gzipped FASTQ files directly. However, since pigz is used currently in a separate compression step, perhaps it would actually be slower to let SamToFastq do the compression.

Alternatively, to avoid the intermediate uncompressed FASTQs, test whether it is possible to let SamToFastq write to two FIFOs and let pigz read from the FIFOs and write to the final output.

PR #147, opened 2019-11-20
Make SamToFastq write gzip directly

pontushojer:
Added pipes to `bam_to_fastq`which now streams the uncompressed files directly to `pigz`. 

Solves #117

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -76,7 +76,7 @@ def test_duplicate_markers(tmpdir, duplicate_marker):
     assert n_input_fastq_reads <= count_bam_alignments(workdir / "mapped.sorted.tag.mkdup.bam")
 
 
-def test_final(tmpdir):
+def test_final_compressed_reads_exist(tmpdir):
     workdir = tmpdir / "analysis"
     init(workdir, TESTDATA_READS)
     copy_config(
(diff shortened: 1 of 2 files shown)


marcelm:
This is a very nice solution! I haven’t used `pipe()` before, so I would have tried to create the FIFO manually. But this is better.

However, I’d like to point out that there are no tests for the `reads.1.final.fastq.gz` and `reads.2.final.fastq.gz` files, so I cannot actually tell from the Travis results whether this modification actually works.

So perhaps this would be a good time to add a test along with this change. The test could at the moment simple ensure that the file exists and is not empty.

pontushojer:
I added a test to see if the final files have been created now. 

The snakemake `pipe()` function is nice. I have however have problems implementing it earlier for programs that run in parallel. Snakemake does not scale the number of threads between the piped rules. But in this case it is easy to apply :)

tests/test_cli.py
```diff
@@ -74,3 +74,18 @@ def test_duplicate_markers(tmpdir, duplicate_marker):
     run(workdir=workdir, targets=["mapped.sorted.tag.mkdup.bam"])
     n_input_fastq_reads = 2 * count_fastq_reads(Path(workdir / "trimmed_barcoded.1.fastq.gz"))
     assert n_input_fastq_reads <= count_bam_alignments(workdir / "mapped.sorted.tag.mkdup.bam")
+
+
+def test_final(tmpdir):
```
marcelm: It’s helpful to use long, explanatory function names in these tests. You could name this `test_final_compressed_reads_exist` or similar. You’ll then also see quicker what a test is supposed to test when pytest tells you the name of it on failure.

pontushojer: That is absolutely true! Will change.

FrickTobias:
I have some things I noticed in the pytest file, but I'll create a separate PR for those.

PR #148, opened 2019-11-20
Pytest update

FrickTobias:
I noticed some inconsistencies in the pytest file when I was about to add some other things so I thought I'd make some changes.

- I moved the test config file and the genome reference to the top to avoid having repetitive hard-coded strings for the same thing at multiple locations in the script.
- The `copy_config` documentation was updated.
- The `copy_config` didn't use the corresponding input argument for `genome_reference` to the function, now it does.
- I removed the `read_mapper` input from test which does not run several different ones.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -7,7 +7,8 @@ from blr.cli.init import init
 from blr.cli.run import run
 
 TESTDATA_READS = Path("testdata/reads.1.fastq.gz")
-
+TEST_CONFIG = Path("tests/test_config.yaml")
+REFERENCE_GENOME = "testdata/chr1mini.fasta"
 
 def count_bam_alignments(path):
     with pysam.AlignmentFile(path) as af:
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -26,13 +28,13 @@ def count_fastq_reads(path):
...
-    """Copy config, possibly changing genome_reference or read_mapper"""
+    """Copy config, possibly changing any non-None arguments"""
 
     with open(source) as infile:
         with open(target, "w") as outfile:
             for line in infile:
                 if genome_reference is not None and line.startswith("genome_reference:"):
-                    path = Path("testdata/chr1mini.fasta").absolute()
+                    path = Path(genome_reference).absolute()
                     line = f"genome_reference: {path}\n"
```
pontushojer: You could remove the `path` variable and just use `genome_reference` directly in the f-string.  There is no need to make it into a absolute `Path` object, as this has already been done on assignment.

marcelm:
I think this looks good in principle.

Note that, while I haven’t mentioned this before, I think `copy_config` should be replaced with something more robust in the medium term. For example, we have an [`igdiscover config`](https://github.com/NBISweden/IgDiscover/blob/master/src/igdiscover/cli/config.py) command in IgDiscover that uses `ruamel.yaml` to modify the YAML configuration file. Better than line-by-line search for the correct key.

tests/test_cli.py
FrickTobias: I was also thinking it might be better to remove the `genome_reference` as input entirely and instead just convert the `test_config.yaml` default for `genome_reference` to it's absolute path. Then we can't change reference but we won't need to specify it in the script nor give it as input for all test functions which involve `run`. 

What do you think about that?

FrickTobias: Hm, I ran in some trouble with the different paths required from the pytest scripts and the bash script so I went with your suggestion

PR #151, opened 2019-11-21
Removed pipes due to sporadic errors

pontushojer:
After merging #147 we got sporadic errors where the named pipe could not be found, example:

```
/bin/bash: reads.1.final.fastq: No such file or directory
```

We reverted to making picard SamToFastq write gzipped file directly for now.


FrickTobias:
I was testing this with him so I agree.

PR #153, opened 2019-11-21
Add bamtofq & remove picardtools

FrickTobias:
By writing our own script we can customize it to our needs and since we are only using picardtools for this step we can get rid of that dependency.

Currently it has two benefits:
- It does not add trailing `/1` and `/2` to the output reads (which causes errors in a lot of fastq handling software since the headers doesn't match).
- It can now be set to add a BAM tag to the header in the output fastq files (which otherwise has to be re-tagged before putting into wfa2tenx & Longranger).


src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
+"Writes fq files from bam files"
+
+import logging
+import pysam
+
```
pontushojer:
```suggestion
```
Remove newline

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
+"Writes fq files from bam files"
```
pontushojer:
```suggestion
"Writes paired or interleaved FASTQ from BAM file."
```

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+from blr import utils
+
+logger = logging.getLogger(__name__)
+
+
+def main(args):
+    logger.info(f"Starting analysis")
+    query_cache = dict()
+    summary = Summary()
+    out_interleaved = not args.r2_out
```
pontushojer: The interleaved options does not work as currently implemented. `r2_out`is a required positional argument. I suggest dropping interleaved FASTQ compatibility for now.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+            summary.tot_reads_in += 1
+
+            # Read pair cache system
+            if query.query_name not in query_cache:
+                query_cache[query.query_name] = query
+                continue
+            else:
+                query_1 = query_cache[query.query_name]
+                query_2 = query
+                del query_cache[query.query_name]
```
pontushojer: You can remove this line and instead write.

```
query_1 = query_cache.pop(query.query_name)
```

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+    :param query_1: paired pysam alignment object
+    :param query_2: paired pysam alignment object
+    :return: read1 and read2 as pysam alignment bjects
+    """
+    if query_1.is_read1 and query_2.is_read2:
+        read1, read2 = query_1, query_2
+    elif query_1.is_read2 and query_2.is_read1:
+        read1, read2 = query_2, query_1
+    else:
+        read1, read2 = None, None
```
pontushojer: In what condition is this necessary? Is there ever a case where the queries will not be read1 or read2?

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+
+def turn_readpair(query_1, query_2):
+    """
+    Finds and set which BAM alignment is read 1 and read 2. Returns None if not read 1 and read 2 found.
+    :param query_1: paired pysam alignment object
+    :param query_2: paired pysam alignment object
+    :return: read1 and read2 as pysam alignment bjects
+    """
+    if query_1.is_read1 and query_2.is_read2:
+        read1, read2 = query_1, query_2
```
pontushojer:
```suggestion
        return = query_1, query_2
```
I am a fan of returning early if possible, this saves variable assignment. This function has that opportunity.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+    def print_stats(self):
+        for object_variable, value in vars(self).items():
+            logger.info(f"{object_variable}: {value}")
+
+
+def add_arguments(parser):
+    parser.add_argument("bam", help="BAM file input")
+    parser.add_argument("r1_out", help="Name for r1 .fq output")
+    parser.add_argument("r2_out", help="Name for r2 .fq output")
+    parser.add_argument("-t", "--tag", type=str, metavar="<STRING>", help="BAM tag to add to header in fq out.")
```
pontushojer:
```suggestion
    parser.add_argument("-t", "--tag", type=str, metavar="<STRING>", help="BAM tag to add to header in fq out.")

```
Missing newline at end of script.

src/blr/cli/bamtofq.py
FrickTobias: I copied it from an old script so it might have had a corner case it was fixing but since I can't remember it I'll remove it for now.

src/blr/cli/bamtofq.py
FrickTobias: Yeah, that's cleaner

src/blr/cli/bamtofq.py
pontushojer: Also there is a looooot of different names starting with query in this section. I find it a little confusing. Possibly rename:
query_cache --> pair_cache
query_1 --> mate
query_2 --> query (i.e don't rename).

src/blr/cli/bamtofq.py
FrickTobias: Yeah I just forgot to remove this line.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+from tqdm import tqdm
+import dnaio
+
+from blr import utils
+
+logger = logging.getLogger(__name__)
+
+
+def main(args):
+    logger.info(f"Starting analysis")
```
pontushojer:
```suggestion
    logger.info("Starting analysis")
```
This string does not have to be f-string.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,96 @@
...
+            r1, r2 = turn_readpair(query_1, query_2)
+
+            r1_as_dnaio_object = dnaio._core.Sequence(name=r1.query_name, sequence=r1.seq, qualities=r1.qual)
+            r2_as_dnaio_object = dnaio._core.Sequence(name=r2.query_name, sequence=r2.seq, qualities=r2.qual)
+
+            writer.write(r1_as_dnaio_object, r2_as_dnaio_object)
+            summary.tot_reads_out += 2
+
+    summary.print_stats()
+    logger.info(f"Finished")
```
pontushojer:
```suggestion
    logger.info("Finished")
```
This string does not have to be f-string.

src/blr/cli/bamtofq.py
FrickTobias: Since it doesn't write interleaved I'm skipping this.

src/blr/cli/bamtofq.py
FrickTobias: Can't say I have a strong opinion so I'll abide.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,88 @@
+"Writes fq files from BAM files"
```
pontushojer:
```suggestion
"Writes paired FASTQ files from BAM files"
```

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def print_stats(self):
+        for object_variable, value in vars(self).items():
+            logger.info(f"{object_variable}: {value}")
+
+
+def add_arguments(parser):
+    parser.add_argument("bam", help="BAM file input")
+    parser.add_argument("r1_out", help="Name for r1 .fq output")
+    parser.add_argument("r2_out", help="Name for r2 .fq output")
```
pontushojer: One final comment about the naming here. 

In the other scripts I have implemented a somewhat common naming regime where the the input file is named `input` and the output file `output` (in this case `output1` and `output2`). What fo you think about implementing the same here?

Also the help message for `r1_out` and `r2_out` should be changed to something like: "Read1/2 FASTQ output".

pontushojer:
Have you run any runtime comparison between picards and your tools? It would be nice to see how they compare as far as performance.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,93 @@
+"Writes paired fq files from BAM files"
```
pontushojer:
```suggestion
"Writes paired FASTQ files from BAM files"
```
This got changed back for some reason

FrickTobias:
So I compared picard and my script in two cases, one with 128 Gb memory and one with 6.2 Gb memory for a BAM file with 670 M reads.

So current `blr bamtofq` seems to
- be faster
- shut down if more memory is required (rather than not doing anything and keep running which picard did).
- require more cpu processing power
- be way more memory efficient

### Commands used:
```
blr bamtofq \
	map.sort.tag.rmdup.x2.filt.bam \
	-o $SNIC_TMP/r1.fq.gz \
	-p $SNIC_TMP/r2.fq.gz
```
```
picard -Xms90g SamToFastq \
	I=map.sort.tag.rmdup.x2.filt.bam \
	FASTQ=$SNIC_TMP/r1.fq.gz \
	SECOND_END_FASTQ=$SNIC_TMP/r2.fq.gz
```
### blr bamtofq
**6.2 Gb:** 

rackham-snic2019-3-583-tobiasf-10923481


**128 Gb:**  
rackham-snic2019-3-583-tobiasf-10923482

### picard SamToFastq
**6.2 Gb:** 
rackham-snic2019-3-583-tobiasf-10923483

**128 Gb:**  
rackham-snic2019-3-583-tobiasf-10923484

pontushojer:
Great results, seem that `bamtofastq` is better than picard in most ways. One thing, picard (java) is here limited to one thread. You can change this by adding `-XX:ParallelGCThreads=10` or whatever number you want. I don't think it will make much difference in this case though.

src/blr/cli/bamtofq.py
```diff
@@ -0,0 +1,93 @@
...
+
+            # Fetch and add bam tag to header (if not found header is not modified)
+            if args.tags:
+                add_tag_to_header(mate, query, args.tags)
+
+            # Find out which query is read 1 and read 2 (so they will be output to correct file)
+            r1, r2 = find_r1_r2(mate, query)
+
+            # Write output
+            r1_as_dnaio_object = dnaio._core.Sequence(name=r1.query_name + " 1", sequence=r1.seq, qualities=r1.qual)
```
marcelm: Please use `dnaio.Sequence`

marcelm:
While I think it’s fun to write one’s own tool, have you considered some of the alternatives? There’s `samtools fastq` for example. It has lots of options, which could make it work the way you want. It has `-n`, which will make it omit the `/1` and `/2`, and it has `-T`, which can be used to copy BAM tags to the FASTQ header.

FrickTobias:
I didn't know about `samtools fastq` (nor that it could transfer the BAM tags) when I wrote this and I already had this script lying around in an old scripts folder. 

Currently, since we don't have any multiprocessing implemented, `samtools` might be a better solution to this. That would also be a more standardized solution for future readers, so maybe it would be worth using that instead as of now?

The advantages I see using `bamtofq` is we can specify exactly what kind of format we want but I can't think of something we might need more than what `samtools` is offering. What do you guys think?

FrickTobias:
Closing this PR in favor of #161

Issue #79, opened 2019-09-09
Common summary class


pontushojer:
Issue from discussion in #76 

>I am also thinking if we should maybe have a standard `summary` class for these scripts that we import into all of them. This could maybe use something other than the predefined variables used currently. I like using the `collections.Counter`for most of these things, it uses a dictionary to store these instead of having to define separate variables.

PR #156, opened 2019-11-22
Add common summary tool

pontushojer:
Related to #79.

I removed the different `Summary` classes used in the scripts `buildmolecules`, `filterclusters`and `clusterrmdup` and instead used `collecitons.Counter` to store these stats in a dictionary. This makes it easy to add new statistics to the summary as you define them on the spot rather then in the `init` function of the `Summary` class. 

**Examples**
To add a new counter just do something like:

```
for thing in things:
   if thing.is_nice():
      summary["My counts"] += 1
```
You can also add values that you want in the summary directly e.g

```
my_list = [0, 2, 51, 64, 621, 2, 51]
summary["My sum"] = sum(my_list)
```

I added a common `print_stats` function in `utils` that can produce nicely formatted statistics from a dictionary-like object such as the `Counter`. It will produce output like this:

```
===========================================
STATS SUMMARY - blr.cli.my_tool
-------------------------------------------
My counts                            2,316
My sum                                 725
===========================================
```

### DIFF
--- a/src/blr/cli/buildmolecules.py
+++ b/src/blr/cli/buildmolecules.py
@@ -7,23 +7,21 @@ a maximum distance of --window between any given reads.
 
 import pysam
 import logging
-
+from collections import Counter
 from tqdm import tqdm
 
 from blr import utils
 
 logger = logging.getLogger(__name__)
+summary = Counter()
 
 
 def main(args):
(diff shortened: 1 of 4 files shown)


FrickTobias:
I think this looks nice, merging.

Issue #150, opened 2019-11-21
Replace copy_config in test_cli


pontushojer:
It was suggested in #148 that the `copy_config` in `tests/test_cli.py` should be replaced with something more robust.

One options was to do something similar to the [`IgDiscover config`](https://github.com/NBISweden/IgDiscover/blob/master/src/igdiscover/cli/config.py) command.
This uses the [ruamel.yaml](https://yaml.readthedocs.io/en/latest/) module to modify the YAML configuration file.

PR #158, opened 2019-11-26
Added config tool to replace copy_config in test_cli

pontushojer:
See #150 

New command line tool added `blr config`. This is based on the similar tool in [IgDiscover](https://github.com/NBISweden/IgDiscover/blob/master/src/igdiscover/cli/config.py). I further implemented the `validate` function from `snakemake.utils` to compare the changes in the config to the `config.schema.yaml` specifications.  The function expects a file `blr.yaml` in the current directory but this can also be specified as a path using the `--file` option. If the new changes are valid the config file will be overwritten with the new settings. 

**Example: To view current configs leave out `--set` argument.**
```bash
$ blr config
SETTINGS FOR: config set: []
 file: blr.yaml
--- CONFIGS IN: blr.yaml ---
heap_space: 90
cluster_tag: BX    # Used to store barcode cluster id in bam file. 'BX' is 10x genomic default
molecule_tag: MI  # Used to store molecule ID, same as 10x default.
num_mol_tag: MN # Used to store number of molecules per barcode
sequence_tag: RX    # Used to store original barcode sequence in bam file. 'RX' is 10x genomic default
genome_reference:                            # Path to indexed reference
read_mapper: bwa      # Choose bwa, bowtie2 or minimap2
duplicate_marker: sambamba # Choose sambamba or samblaster

#################
# Trim settings #
#################
h1: CAGTTGATCATCAGCAGGTAATCTGG # h1 adaptor sequnce, appears before barcode in read1
h2: CATGACCTCTTGGAACTGTCAGATGTGTATAAGAGACAG # h2 adaptor sequence, appears after barcode in read1
h3: CTGTCTCTTATACACATCT # h3 adaptor sequence, appears after genomics insert in read1 and read2.
barcode_len: 20 # Length of barcode sequence

####################
# Phasing settings #
####################
reference_variants: ../testdata/HG002_GRCh38_GIAB_highconf.chr1mini.vcf #Path to reference variants used for phasing, if not provided then variant will be called by freebayes
phasing_ground_truth: #Path to phased variants used for calculating stats
```

**Example: Change 'barcode_len' from 20 to 10.**
```bash
$ blr config --set barcode_len 10
SETTINGS FOR: config set: [['barcode_len', '10']]
 file: blr.yaml
config - INFO: Changing value of 'barcode_len'': 20 --> 10.
```

**Example: Remove value for reference_variants i.e. change for empty value.**
Note that `None` will not work here as it will change the value to the string `"None"`.
```bash
$ blr config --set reference_variants null
SETTINGS FOR: config set: [['reference_variants', 'null']]
 file: blr.yaml
config - INFO: Changing value of 'reference_variants': ../testdata/HG002_GRCh38_GIAB_highconf.chr1mini.vcf --> None.
````

### CODE REVIEW

File: src/blr/cli/config.py [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+    :return: (data, yaml).
+    """
+    with open(filename) as file:
+        yaml = YAML()
+        data = yaml.load(file)
+    return data, yaml
+
+
+def add_arguments(parser):
+    parser.add_argument("--set", nargs=2, default=[], metavar=("KEY", "VALUE"), action="append",
```
FrickTobias: This `default=[]` doesn't do anything right? If it isn't defined `main` will just write config to `stdout` right?

environment.yml
```diff
@@ -23,3 +23,4 @@ dependencies:
   - tqdm
   - importlib_resources
   - freebayes
+  - ruamel.yaml
```
marcelm: a newline at the end of the file is missing

src/blr/cli/config.py
marcelm: This is just copied from IgDiscover and can be removed, yes. I assumed that it’s necessary for `action="append"`, but it’s not – argparse is smart enough.

marcelm:
> Since we're always copying the config file when using this script I think it makes more sense to abide by having a standard input/output

The file is currently copied, but I would actually suggest not to copy it in the future. For the tests, I think it would be a good model to have `blr init` create a default config file and modify that. That way, we have to maintain only one the one default configuration file and only need to worry about the things that we actually want to change in the tests.

Also, I think `blr init` could actually become a user-facing program (in contrast to `blr tagbam` for example, which we talked about in another PR) that users could use in their scripts to automate running a pipeline. In that case, it makes also more sense to modify `blr.yaml` in place.

pontushojer:
> * **Add:** Functionality to change multiple values. This would open the possibility to use it in `blr init` and we can remove the test_config file.

This is currently implemented. See the following example:

```bash
$ blr config --set barcode_len 10 --set genome_reference /path/to/genome.fasta
SETTINGS FOR: config set: [['barcode_len', '10'],['genome_reference', '/path/to/genome.fasta']]
 file: blr.yaml
config - INFO: Changing value of 'barcode_len'': 20 --> 10.
config - INFO: Changing value of 'genome_reference'': None --> /path/to/genome.fasta.
```

I will also remove the copying of the `test_config.yaml` file and instead edit the provided`blr.yaml` inplace.

FrickTobias:
> The file is currently copied, but I would actually suggest not to copy it in the future. For the tests, I think it would be a good model to have `blr init` create a default config file and modify that. That way, we have to maintain only one the one default configuration file and only need to worry about the things that we actually want to change in the tests.

I don't really see a big difference by having a default config being written every time a function is run compared to copying a default config. However, to me it makes more sense to not rewrite files if it isn't necessary. Either way both solutions could be made to modify any non-default values when writing the new file which would avoid writing the file numerous times.

> Also, I think `blr init` could actually become a user-facing program (in contrast to `blr tagbam` for example, which we talked about in another PR) that users could use in their scripts to automate running a pipeline. In that case, it makes also more sense to modify `blr.yaml` in place.

So I'm suggesting combining the modifying into `blr init`. This could then be run without supplying any further arguments and have to be modified manually like you are suggesting, but also be very easy to automate.

```
blr init --reads r1.fq.gz outdir --genome-reference hg38.fa --cluster-dist 1 <other config modifications>
blr run --dir outdir 
```
or
```
for r in my_sample_list;
do 
    blr init --reads $r outdir_$r --genome-reference my/refs/hg38.fa <other config modifications>
    blr run
done
```

FrickTobias:
I talked to Pontus about it and from thinking about rerunning the analysis from e.g. halfway I agree having a modifying script makes sense.

Issue #154, opened 2019-11-22
Compress barcodes.fasta after use.


pontushojer:
`barcodes.fasta` containing the uncorrected barcodes is now left uncompressed after the workflow. In the latest dataset this took up ~31Gb of space. After it has been used for clustering and tagging the FASTQ it should therefore be compressed to save space.

pontushojer:
Optionally we could create a compressed `barcodes.fasta.gz` and then stream the file into `starcode` with something like:

```bash
starcode -i <(pigz < barcodes.fasta.gz) -o barcodes.clstr ...
```

`blr tagfastq` is already compatible with gzipped files.

PR #160, opened 2019-11-27
Compress barcode.fasta

FrickTobias:
Solves #154. Everything was compatible with reading .gz or from stdin.

### DIFF
--- a/src/blr/rules/trim.smk
+++ b/src/blr/rules/trim.smk
@@ -62,7 +62,7 @@ rule trim_and_tag:
 rule extract_DBS:
     # Extract barcode sequence from read1 FASTQ
     output:
-        fastq="barcodes.fasta.gz",
+        fastq="barcodes.fasta.gz"
     input:
          fastq="reads.1.fastq.gz"
     log: "cutadapt_extract_DBS.log"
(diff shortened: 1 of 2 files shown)


src/blr/rules/trim.smk
```diff
@@ -62,7 +62,7 @@ rule trim_and_tag:
 rule extract_DBS:
     # Extract barcode sequence from read1 FASTQ
     output:
-        fastq="barcodes.fasta",
+        fastq="barcodes.fasta.gz",
```
pontushojer:
```suggestion
        fastq="barcodes.fasta.gz"
```
Seems there is a rouge comma here.

src/blr/rules/trim.smk
```diff
@@ -62,7 +62,7 @@ rule trim_and_tag:
...
     input:
         fastq="reads.1.fastq.gz",
```
pontushojer:
```suggestion
        fastq="reads.1.fastq.gz"
```
Also here

pontushojer:
Short and sweet! 👍 

I will merge once the comments are handled.

FrickTobias:
Hm, don't know why we had the commas there in the first place but I accepted your suggestions.

marcelm:
It was probably me who added the trailing comma out of habit because I’ve started to do so in regular Python. See how [Black does it](https://test-black.readthedocs.io/en/style-guide/the_black_code_style.html#trailing-commas). In short, having trailing commas on every line in multi-line expressions reduces the diff when you add a new expression. But it’s totally fine to remove it here because we’re not being consistent anyway, so no worries.

PR #161, opened 2019-11-27
Replace picard SamToFastq with samtools fastq

FrickTobias:
See discussion in #153. 

**TL;DR:** We can remove picardtools and also get the compatibility to
- add bam tags to header
- use threading
- get less dependencies
- skip getting /1 and /2 written to the end of the header (causes errors in most software)
- probably have more efficient RAM usage, however this is not tested for samtools fastq


src/blr/Snakefile
```diff
@@ -168,12 +168,13 @@ rule bam_to_fastq:
...
         bam = "mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam"
-    log: "picard_samtofastq.log"
+    log: "samtools-fastq.log"
     shell:
-        "picard -Xms{config[heap_space]}g SamToFastq"
-        " I={input.bam}"
-        " FASTQ={output.r1_fastq}"
-        " SECOND_END_FASTQ={output.r2_fastq} 2>> {log}"
+        "samtools fastq"
+        " -@ {threads}"
```
pontushojer: For `threads` to be useful we need to define the maximum number of threads for this rule. Add `threads: 20` after `log: "samtools-fastq.log"`

FrickTobias: I just remembered, it's fixed now.

pontushojer:
One thing, we are getting this message in the test file.

```
+picard SamToFastq --version
tests/run.sh: line 7: picard: command not found
```
You need to remove the line `picard SamToFastq --version` from `tests/run.sh`.

FrickTobias:
Remembered there are also heap space and other arguments that can be removed. Hold tight.

FrickTobias:
Now everything should be removed, sorry about that!

marcelm:
Wow, very nice. I like how everything gets shorter, in particular the list of dependencies required on Linux!

PR #163, opened 2019-11-28
Changed VCF file naming

pontushojer:
It has been hard to separate between called variants and those supplied from a reference as both were referred to as `reference.vcf`. I have now change to separate these so that we have:

- `variants.reference.vcf`: which is a symlink to a reference vcf file.
- `variants.called.vcf`: which contains the called variants, currently from `freebayes`.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -99,4 +99,4 @@ def test_call_variants(tmpdir):
     )
     target = "variants.called.vcf"
     run(workdir=workdir, targets=[target])
-    assert Path(workdir / target).exists()
+    assert Path(workdir / target).is_file()
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -78,17 +78,25 @@ def test_final_compressed_reads_exist(tmpdir):
         assert Path(workdir / filename).exists()
 
 
-@pytest.mark.parametrize("reference_variants", [REFERENCE_VARIANTS, "null"])
-def test_reference_variants(tmpdir, reference_variants):
+def test_link_reference_variants(tmpdir):
     workdir = tmpdir / "analysis"
     init(workdir, TESTDATA_READS)
     change_config(
         workdir / DEFAULT_CONFIG,
-        [("genome_reference", REFERENCE_GENOME), ("reference_variants", reference_variants)]
+        [("genome_reference", REFERENCE_GENOME), ("reference_variants", REFERENCE_VARIANTS)]
     )
-    target = "reference.vcf"
-    run(workdir=workdir, targets=["reference.vcf"])
-    if reference_variants != "null":
-        assert Path(workdir / target).is_symlink()
-    else:
-        assert Path(workdir / target).is_file()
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -78,17 +78,25 @@ def test_final_compressed_reads_exist(tmpdir):
...
+def test_call_variants(tmpdir):
+    workdir = tmpdir / "analysis"
+    init(workdir, TESTDATA_READS)
+    change_config(
+        workdir / DEFAULT_CONFIG,
+        [("genome_reference", REFERENCE_GENOME), ("reference_variants", "null")]
+    )
+    target = "variants.called.vcf"
+    run(workdir=workdir, targets=[target])
+    assert Path(workdir / target).exists()
```
FrickTobias: What was the benefit of separating these two? 

I'm thinking we could just modify `target` instead as you do in `phasing.smk`, something like:
```
@pytest.mark.parametrize("reference_variants", [REFERENCE_VARIANTS, "null"])
def test_reference_variants(tmpdir, reference_variants):
    workdir = tmpdir / "analysis"
    init(workdir, TESTDATA_READS)
    change_config(
        workdir / DEFAULT_CONFIG,
        [("genome_reference", REFERENCE_GENOME), ("reference_variants", reference_variants)]
    )
    target =  "variants.reference.vcf" if config["reference_variants"] else "variants.called.vcf"
    run(workdir=workdir, targets=target)
    if reference_variants != "null":
        assert Path(workdir / target).is_symlink()
    else:
        assert Path(workdir / target).is_file()
```

FrickTobias: 102: `Path(workdir / target).is_file()`

pontushojer: I changed to two separate because I felt it was hard tell from the test what was actually being tested. `test_reference_variants` or as you suggest here `test_link_and_call_variants` tests for two separate things so therefore it seems more natural to keep them separate. Currently the tests state:
```bash
tests/test_cli.py::test_reference_variants[/home/travis/build/FrickTobias/BLR/testdata/HG002_GRCh38_GIAB_highconf.chr1mini.vcf] PASSED [ 90%]
tests/test_cli.py::test_reference_variants[null] PASSED                  [100%]
```

This is a bit hard to interpret since the only difference in input does not tell what is tested. 

One other benefit with splitting is if we want to test multiple variant callers, then we can easily parameterize them in these split tests.

pontushojer: I will change the `exists()` for `is_file()`.

FrickTobias: Makes sense for future purposes. However I think we should rename the test to `test_variant_callers` to indicate the rule is supposed to be used for multiple variant callers when we add them.

pontushojer: I think we can make that change when we actually have other callers, right now it does not make much sense.

FrickTobias: I'm not sure if I agree but since it's so minor I'll go ahead and merge.

PR #165, opened 2019-11-29
Update tagbam

FrickTobias:
To accommodate mappers like `ema` we need to have some more flexibility to `tagbam` since it requires, and outputs different output format of headers.

I rewrote `tagbam` to so it searches for regex patterns in query headers for any number of specified SAM tags. More importantly it is written to be compatible with adding new regex functions to search for other patterns. 

To add a new patten you you need to:
- write regex function (equivalent to build_regex_sam_tag())
- add format to options in argparser
- link format & regex function in REGEX_FUNCTION_DICT

**In the future** we could also some more modes for different mappers copying different information to tags.
- `--safe-mode` If tag already exists, check the header tag was identical (or cause error)
- `--soft-mode` Don't overwrite if tag already has a value.

**Bonus:** Since it is using regex rather than split() it now works regardless of what separator is used between the read name and barcode (currently `_`) and tags can be in any order.

**Lastly** I made `samtools fastq` include tags into the final fastq files.

### DIFF
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -45,7 +45,7 @@ def main(args):
                         read.set_tag(tag, match.group("value"), value_type=match.group("type"))
                     summary[f"Reads with tag {tag}"] += 1
                 else:
-                    summary[f"reads without tag {tag}"] += 1
+                    summary[f"Reads without tag {tag}"] += 1
 
             out.write(read)
 
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -6,6 +6,9 @@ import pysam
 import logging
 import re
 from tqdm import tqdm
+from collections import Counter
+
+from blr.utils import print_stats
 
 logger = logging.getLogger(__name__)
 
@@ -15,6 +18,7 @@ ALLOWED_BAM_TAG_TYPES = "ABfHiZ" # From SAM format specs https://samtools.github
 def main(args):
     logger.info("Starting analysis")
     alignments_missing_bc = 0
+    summary = Counter()
 
     # Read SAM/BAM files and transfer barcode information from alignment name to SAM tag
     with pysam.AlignmentFile(args.input, "rb") as infile, \
@@ -31,23 +35,20 @@ def main(args):
                     divider = read.query_name[match.start()-1]
                     read.query_name = read.query_name.replace(divider + full_tag_string, "")
                     read.set_tag(match.group("tag"), match.group("value"), value_type=match.group("type"))
-
-                    # summary: add +1 to reads with [tag]
+                    summary[f"reads with tag {tag}"] += 1
 
             out.write(read)
 
-    logger.info(f"Alignments missing barcodes: {alignments_missing_bc}")
+    print_stats(summary, name="stats")
     logger.info("Finished")
 
 
 def find_tag(header, bam_tag, allowed_value_chars="ATGCN"):
     """
-
-    :param header: pysam header
-    :param bam_tag: SAM tag to search for
-    :param allowed_value_chars: characters allowed in SAM value
-    :param value_length: max length of SAM value
-    :param len_variance:
+    Finds BAM tags in header and returns regex match objects.
+    :param header: strm pysam header
+    :param bam_tag: str, SAM tag to search for
+    :param allowed_value_chars: str, characters allowed in SAM value
     :return:
     """
 
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -43,7 +43,7 @@ def main(args):
                     read.query_name = read.query_name.replace(divider + full_tag_string, "")
                     if not args.only_remove:
                         read.set_tag(tag, match.group("value"), value_type=match.group("type"))
-                    summary[f"reads with tag {tag}"] += 1
+                    summary[f"Reads with tag {tag}"] += 1
                 else:
                     summary[f"reads without tag {tag}"] += 1
 
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -1,5 +1,6 @@
 """
-Transfers tags from query headers to SAM tags. Currently tags in header must follow SAM tag format, e.g. BC:Z:<SEQUENCE>.
+Transfers SAM tags from query headers to SAM tags. Currently tags in header must follow SAM tag format, e.g.
+BC:Z:<SEQUENCE>.
 """
 
 import pysam
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -49,7 +49,7 @@ def main(args):
 
             out.write(read)
 
-    print_stats(summary, name="stats")
+    print_stats(summary, name=__name__)
     logger.info("Finished")
 
 
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -12,12 +12,11 @@ from blr.utils import print_stats
 
 logger = logging.getLogger(__name__)
 
-ALLOWED_BAM_TAG_TYPES = "ABfHiZ" # From SAM format specs https://samtools.github.io/hts-specs/SAMtags.pdf
+ALLOWED_BAM_TAG_TYPES = "ABfHiZ"  # From SAM format specs https://samtools.github.io/hts-specs/SAMtags.pdf
 
 
 def main(args):
     logger.info("Starting analysis")
-    alignments_missing_bc = 0
     summary = Counter()
 
     # Read SAM/BAM files and transfer barcode information from alignment name to SAM tag
@@ -32,7 +31,7 @@ def main(args):
                 # If tag string is found, remove from header and set SAM tag value
                 if match:
                     full_tag_string = match.group(0)
-                    divider = read.query_name[match.start()-1]
+                    divider = read.query_name[match.start() - 1]
                     read.query_name = read.query_name.replace(divider + full_tag_string, "")
                     read.set_tag(match.group("tag"), match.group("value"), value_type=match.group("type"))
                     summary[f"reads with tag {tag}"] += 1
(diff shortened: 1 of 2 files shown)


src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
+    """
+    Formats a string to fit the regex pattern for getting named match objects groups
+    :param pattern: str, regex pattern
+    :param name: str, name of the match group
+    :return: str, named regex string
+    """
+    prefix = "(?P<"
+    infix = ">"
+    suffix = ")"
+    return prefix + name + infix + pattern + suffix
```
pontushojer:
```suggestion
    return "".join([prefix, name, infix, pattern, suffix])
```
`join` is faster then adding strings together. When you add strings it sets the new string for each addition i.e 
first `prefix + name`, 
then `prefix + name + infix`, 
then `prefix + name + infix + pattern`, 
then `prefix + name + infix + pattern + suffix`

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
 
     parser.add_argument("-o", "--output", default="-",
                         help="Write output BAM to file rather then stdout.")
-    parser.add_argument("-b", "--barcode-tag", default="BX",
-                        help="SAM tag for storing the error corrected barcode. Default: %(default)s")
-    parser.add_argument("-s", "--sequence-tag", default="RX",
-                        help="SAM tag for storing the uncorrected barcode sequence. Default: %(default)s")
+    parser.add_argument("-t", "--tags", default=["BX"], nargs="*", help="List of SAM tags. Default: %(default)s")
+    parser.add_argument("--only-remove", action="store_true", help="Only remove tag from header, will set SAM tag.")
+    parser.add_argument("--pattern-type", default="sam", choices=["sam"],
```
pontushojer: Why have a argument that only has one option?

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
 def main(args):
+    # Can't be at top since function are defined later
+    REGEX_FUNCTION_DICT = {
+        "sam": build_regex_sam_tag
+    }
 
-    # Generate dict with bc => bc_cluster consensus sequence
     logger.info("Starting analysis")
-    alignments_missing_bc = 0
+    summary = Counter()
```
pontushojer: I usually define this outside the function scope e.g. like `logger`. Then it is accessible and can be updated in any function in the script.

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
-                raw_barcode = raw_barcode_tag.split(":")[-1]
-                read.set_tag(args.sequence_tag, raw_barcode, value_type='Z')
+                # If tag string is found, remove from header and set SAM tag value
+                if match:
+                    full_tag_string = match.group(0)
+                    divider = read.query_name[match.start() - 1]
+                    read.query_name = read.query_name.replace(divider + full_tag_string, "")
+                    if not args.only_remove:
+                        read.set_tag(tag, match.group("value"), value_type=match.group("type"))
+                    summary[f"reads with tag {tag}"] += 1
```
pontushojer:
```suggestion
                    summary[f"Reads with tag {tag}"] += 1
```
In the other script the summary key start with capital letter.

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
+                else:
+                    summary[f"reads without tag {tag}"] += 1
```
pontushojer:
```suggestion
                    summary[f"Reads without tag {tag}"] += 1
```

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
 
             out.write(read)
 
-    logger.info(f"Alignments missing barcodes: {alignments_missing_bc}")
+    print_stats(summary, name="stats")
```
pontushojer:
```suggestion
    print_stats(summary, name=__name__)
```
In the other scripts I have included the script name in the summary table

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
+    pattern_tag_value = f"[{allowed_value_chars}]+"
+    pattern_tag_types = f"[{ALLOWED_SAM_TAG_TYPES}]"
+
+    # Add strings to name match object variables to match.group(<name>)
+    regex_bam_tag = add_regex_name(pattern=bam_tag, name="tag")
+    regex_allowed_types = add_regex_name(pattern=pattern_tag_types, name="type")
+    regex_tag_value = add_regex_name(pattern=pattern_tag_value, name="value")
+
+    # regex pattern search for tag:type:value
+    regex_string = r":".join([regex_bam_tag, regex_allowed_types, regex_tag_value])
```
pontushojer: You could just return this directly without defining a new variable.

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
+                        help="BAM file with SAM tag info in header. To read from stdin use '-'.")
 
     parser.add_argument("-o", "--output", default="-",
                         help="Write output BAM to file rather then stdout.")
-    parser.add_argument("-b", "--barcode-tag", default="BX",
-                        help="SAM tag for storing the error corrected barcode. Default: %(default)s")
-    parser.add_argument("-s", "--sequence-tag", default="RX",
-                        help="SAM tag for storing the uncorrected barcode sequence. Default: %(default)s")
+    parser.add_argument("-t", "--tags", default=["BX"], nargs="*", help="List of SAM tags. Default: %(default)s")
+    parser.add_argument("--only-remove", action="store_true", help="Only remove tag from header, will set SAM tag.")
```
pontushojer: What do expect the use of this to be? Is there anywhere we would need only to remove from the tags?

src/blr/cli/tagbam.py
marcelm: However, globals break encapsulation and you should avoid them whenever you can. For `logger`, it’s practical (and it’s designed to be used that way), but for statistics, it isn’t too hard to keep them local. Try to design the script you’re writing as if you were designing a library. For libraries, global state is evil because it makes them harder to use. You’d get incorrect statistics if you called the `main()` function twice.

marcelm:
Regular expression matching is relatively slow compared to just using `.split()` and friends, so make sure this isn’t degrading performance.

Also, you should use `re.compile()` to pre-compile all regular expressions. Otherwise, there will be an implicit compilation step every time they are used. This may already get the performance to acceptable levels.

src/blr/cli/tagbam.py
FrickTobias: Cool, had no idea!

I'll also include a reference look for method 6 in this [investigation](https://waymoot.org/home/python_string/).

marcelm: The regular expression is created only once, and the runtime is therefore negligible compared to the other work this script does. Also, `''.join()` may not actually be faster because to use it you need to build a list first.

Base your decision on what to use here only on what you find more readable.

src/blr/cli/tagbam.py
FrickTobias: To prepare it for adding more formats since one of the main points was to make it fit for searching for other pattern in the future. It's specified in the script description it **currently** only searches for SAM format tags.

>To accommodate mappers like ema we need to have some more flexibility to tagbam since it requires, and outputs different output format of headers.
>
>I rewrote tagbam to so it searches for regex patterns in query headers for any number of specified SAM tags. More importantly it is written to be compatible with adding new regex functions to search for other patterns.

src/blr/cli/tagbam.py
marcelm: But leaving it in is a way to document what is returned.

src/blr/cli/tagbam.py
marcelm: YAGNI ;-)

src/blr/cli/tagbam.py
pontushojer: I guess it is not too difficult to keep it local. Then we should probably change this in the other script where this is implemented.

src/blr/cli/tagbam.py
pontushojer: Whatever you prefer then ;)

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,100 @@
...
+import re
 from tqdm import tqdm
+from collections import Counter
+
+from blr.utils import print_stats
 
 logger = logging.getLogger(__name__)
 
+ALLOWED_SAM_TAG_TYPES = "ABfHiZ"  # From SAM format specs https://samtools.github.io/hts-specs/SAMtags.pdf
+
```
pontushojer: Here we allow for any SAM tag type, we are however only using "Z" or strings at the moment. I would assume that only including this would speed up the regex search.

src/blr/cli/tagbam.py
FrickTobias: I understand it as `ema` takes input headers like this `@header:<BC SEQ> <BC TAG>` and outputs query headers like this `header:<BC SEQ>` with the `<BC TAG>` info is already set to a tag after mapping.

In that case one should write a new ema pattern and run it with the `--only-remove` flag to get rid of the duplicated information in the header.

The same goes for `bwa mem` if we want to use the `-H` option to tag the BAM file while mapping since it doesn't remove the header info either.

FrickTobias:
> Regular expression matching is relatively slow compared to just using `.split()` and friends, so make sure this isn’t degrading performance.

Sounds reasonable, I'll check it out!

> Also, you should use `re.compile()` to pre-compile all regular expressions. Otherwise, there will be an implicit compilation step every time they are used. This may already get the performance to acceptable levels.

Cool! Haven't used much regex before so didn't know about that, I will implement it.

src/blr/cli/tagbam.py
FrickTobias: > YAGNI ;-)

I see where you are coming from but...

>I understand it as `ema` takes input headers like this `@header:<BC SEQ>` `<BC TAG>` and outputs query headers like this header:`<BC SEQ>` with the `<BC TAG>` info is already set to a tag after mapping.

So in that case we would like to remove the tag & specify it to search for a non-SAM tag format regex pattern.

FrickTobias:
So I did some testing on this and it turns out when reads actually have tags it becomes much slower than the previous method (~1.5-2x the time of the previous depending on exact implementation). 

Therefore I'll close this PR and make a new one based on split where tagbam can fetch read parsing functions based on what mode it is being run in.

Issue #146, opened 2019-11-19
Non-diploid calls


FrickTobias:
From running the current pipeline without a reference `vcf` file I am getting non-diploid calls. These are then causing problems for HapCUT2 when trying to phase them.

However according to the `freebayes` documentation it seems like we should only get diploid samples.

["Call variants assuming a diploid sample"](https://github.com/ekg/freebayes)
```
freebayes -f ref.fa aln.bam >var.vcf
```

Current command used in Snakefile for variant calling

```
"freebayes"
" -f {config[reference]}.fasta"
" {input.bam} 1> {output.vcf} 2> {log}"
```

FrickTobias:
I'm currently trying to run it using `-p 2` to see if it only the default which is not working properly.

marcelm:
How do these calls look in the VCF?

pontushojer:
This is from the error message:

```
ERROR: Non-diploid VCF entry detected. Each VCF entry must have a diploid genotype (GT) field consisting of two alleles in the set {0,1,2} separated by either '/' or '|'. For example, "1/1", "0/1", and "0|2" are valid diploid genotypes for HapCUT2, but "1", "0/3", and "0/0/1" are not.
The invalid entry is: 

[...]

PR #166, opened 2019-11-29
Filter vcf for non-called variants

FrickTobias:
Solves #146 by filtering freebayes output for variants which has no alleles assigned to them.


FrickTobias:
Since it's a minor change at it's been five days I going ahead and merging.

Issue #173, opened 2019-12-06
Clusterrmdup slow for large files. 


pontushojer:
I did a run on three combined datasets (1,145,479,013 read pairs) which run for 72 hours until it was stopped due to overbooked memory. By then it had processed 1,025,783,235 read pairs in the first step of `clusterrmdup`. 

![image](https://user-images.githubusercontent.com/27061883/70311117-49ce7800-1811-11ea-83b7-c342ac74f1ab.png)

I am currently working on speeding it up.

Issue #167, opened 2019-11-29
Write summary as local instances rather than global


FrickTobias:
Try and stay away from globals as much as possible.

_Originally posted by @pontushojer in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDM1MjE2MDQ4Ng==/comments/review_comment_

FrickTobias:
**Update:** The only script left to update is `clusterrmdup`, which will be fixed by PR #174.

PR #174, opened 2019-12-10
clusterrmdup update

pontushojer:
We had some issues with running `clusterrmdup` for large dataset (see #173 ) so I have done some updates to limit memory and increase performance. I also changed some aspects that would cause the script to miss some duplicates based on the [picard definition for duplicates](https://sourceforge.net/p/samtools/mailman/message/25062576/): 

"... find the 5' coordinates and mapping orientations of each read pair.  [...].  It then matches all read pairs that have identical 5' coordinates and orientations and marks as duplicates all but the "best" pair."

The previous script used the 5' and 3' positions of both reads and did not consider the orientation. Thus I have now changed so that only 5' positions and orientation is considered. The image below shows one such scenario were a duplicate between barcodes GTCCTTTGTTCCGTACTAAG and AATAGGTCTTTCGACATACG but the 3' end of read2 is trimmed differently. "test.master.bam" this previous script which miss this duplicate while the new script ("test.new.bam") detects it and merges the barcodes.    

image

I have also had to implement a buffer system to be able to catch all read on a duplicate position. Consider the following scenario.
```
READ PAIR 1:    R1|--------->    <-------|R2
READ PAIR 2:    R1|--------->       <----|R2
Chromosome:  ==================================>
```

Buy the definition stated above the two read pairs (RP) are duplicates but R2 for RP2 is on a different position start position compared to R2 in RP1. Thus it will appear longer down in the coordinate sorted BAM file. The previous script looked at each position separately and would thus drop RP1 as possible duplicate. Now instead each RP within a defined window (currently 500 bp) is considered. Thus these cases can be caught and handled correctly.

**Other changes**
- Move `summary` away from global scope (see #167 )
- Factored out several functions.
- Name changes  

**Runtime for chr22 testdata.** 
Previous script
- real: 10m18.406s, 10m21.591s
- user: 10m16.232s, 10m19.550s
- sys: 0m1.316s, 0m1.035s

This script
- real: 8m3.786s, 8m6.691s
- user: 8m1.998s, 8m5.079s
- sys: 0m1.052s, 0m0.900s

### DIFF
--- a/src/blr/cli/clusterrmdup.py
+++ b/src/blr/cli/clusterrmdup.py
@@ -27,7 +27,8 @@ def main(args):
     pos_prev = 0
     merge_dict = dict()
     buffer_dup_pos = deque()
-    for barcode, read, mate in tqdm(parse_and_filter_pairs(args.input, args.barcode_tag, summary), desc="Reading pairs"):
+    for barcode, read, mate in tqdm(parse_and_filter_pairs(args.input, args.barcode_tag, summary),
+                                    desc="Reading pairs"):
         # Get orientation of read pair
         if mate.is_read1 and read.is_read2:
             orientation = "F"
@@ -182,7 +183,8 @@ def find_barcode_duplicates(positions, buffer_dup_pos, merge_dict, window, summa
     for position in positions.keys():
         tracked_position = positions[position]
         if tracked_position.valid_duplicate_position():
-            summary["Reads at duplicate position"] += tracked_position.reads
+            if not tracked_position.checked:
+                summary["Reads at duplicate position"] += tracked_position.reads
             seed_duplicates(
                 merge_dict=merge_dict,
                 buffer_dup_pos=buffer_dup_pos,
@@ -210,7 +212,7 @@ class PositionTracker:
         self.reads = int()
         self.barcodes = set()
         self.checked = False
-        self.updated = bool()
+        self.updated_since_validation = bool()
         self.read_pos_has_duplicates = bool()
         self.mate_pos_has_duplciates = bool()
 
@@ -222,14 +224,14 @@ class PositionTracker:
         if mate.is_duplicate:
             self.mate_pos_has_duplciates = True
 
-        self.updated = True
+        self.updated_since_validation = True
         self.reads += 2
         self.barcodes.add(barcode)
 
     def valid_duplicate_position(self):
         check = self.read_pos_has_duplicates and self.mate_pos_has_duplciates and len(self.barcodes) >= 2 and \
-                self.updated
-        self.updated = False
+                self.updated_since_validation
+        self.updated_since_validation = False
         return check
 
 
@@ -249,6 +251,11 @@ def seed_duplicates(merge_dict, buffer_dup_pos, position, position_barcodes, win
     # Loop over list to get the positions closest to the analyced position first. When position are out of the window
     # size of the remaining buffer is removed.
     for index, (compared_position, compared_barcodes) in enumerate(buffer_dup_pos):
+
+        # Skip comparison against self.
+        if position == compared_position:
+            continue
+
         compared_position_stop = compared_position[1]
         if compared_position_stop + window >= pos_start_new:
             barcode_intersect = position_barcodes & compared_barcodes
@@ -329,6 +336,6 @@ def add_arguments(parser):
     parser.add_argument("-w", "--window", type=int, default=100000,
                         help="Window size. Duplicate positions within this distance will be used to find cluster "
                              "duplicates. Default: %(default)s")
-    parser.add_argument("--buffer-size", type=int, default=500,
-                        help="Buffer size for collecting duplicates. Note that too low numbers might miss some "
-                             "duplicates while larger numbers increase memory requirement. Default: %(default)s")
+    parser.add_argument("--buffer-size", type=int, default=200,
+                        help="Buffer size for collecting duplicates. Should be around read length. "
+                             "Default: %(default)s")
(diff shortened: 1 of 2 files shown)


src/blr/cli/clusterrmdup.py
```diff
@@ -12,237 +12,297 @@
...
+        barcode = utils.get_bamtag(read, args.barcode_tag)
+        if not barcode:
+            summary["Non tagged reads"] += 2
+            continue
+
+        # Get orientation of read pair
+        if mate.is_read1 and read.is_read2:
+            orientation = "F"
+        else:
+            orientation = "R"
```
FrickTobias: When I read this I first confused this `orientation` with `orientation_ok`, maybe one could specify this as `rp_orientation`?

src/blr/cli/clusterrmdup.py
```diff
@@ -12,237 +12,297 @@
...
-
-            # Wait for mate of read until process
-            if read.query_name not in cache_reads:
-                cache_reads[read.query_name] = read
-                continue
-            else:
-                mate = cache_reads[read.query_name]
-                del cache_reads[read.query_name]
+    buffer_dup_pos = deque()
+    for read, mate in tqdm(parse_filtered_read_pairs(args.input, summary), desc="Reading pairs"):
```
FrickTobias: I think this makes it seem as the function expects a filtered input, how about renaming it as `parse_and_filter_read_pairs`?

src/blr/cli/clusterrmdup.py
```diff
@@ -12,237 +12,297 @@
...
+
+        # Get barcode and confirm that its not None
+        barcode = utils.get_bamtag(read, args.barcode_tag)
+        if not barcode:
+            summary["Non tagged reads"] += 2
+            continue
```
FrickTobias: This should be moved to `meet_requirements()` now when the reader filters reads.

src/blr/cli/clusterrmdup.py
```diff
@@ -12,237 +12,297 @@
...
+        # Based on picard MarkDuplicates definition, see: https://sourceforge.net/p/samtools/mailman/message/25062576/
+        current_position = (mate.reference_start, read.reference_end, orientation)
 
-                    if rp_pos_tuple in current_cache_rp:
-                        current_cache_rp[rp_pos_tuple].add_read_pair_and_bc(read=read, mate=mate, bc=bc_new)
-                    else:
-                        current_cache_rp[rp_pos_tuple] = CacheReadPairTracker(rp_pos_tuple=rp_pos_tuple,
-                                                                              chromosome=chrom_new, read=read,
-                                                                              mate=mate, bc=bc_new)
+        if abs(pos_new - pos_prev) > BUFFER_SIZE or not chrom_new == chrom_prev:
```
FrickTobias:
```suggestion
        if abs(pos_new - pos_prev) > BUFFER_SIZE or chrom_new  not == chrom_prev:
```

FrickTobias: Minor comment: Do we need such a big buffer size? It shouldn't be able to be more than a couple of bp right?

src/blr/cli/clusterrmdup.py
```diff
@@ -12,237 +12,297 @@
...
 
-        self.position_tuple_ID = rp_pos_tuple
-        self.chromosome = chromosome
-        self.current_reads = dict()
+    def __init__(self, position, read, mate, barcode):
+        self.position = position
+        self.reads = int()
         self.barcodes = set()
+        self.checked = False
+        self.updated = bool()
```
FrickTobias:
```suggestion
        self.contains_reads = bool()
```

src/blr/cli/clusterrmdup.py
pontushojer: True

src/blr/cli/clusterrmdup.py
pontushojer: I think `updated` makes more sense here although it may not be obvious. As the position is only removed when its outside the buffer the same position may be evaluated multiple times. I added this so that once it had been evaluated it would only be evaluated again if more reads had been added. Thus is called `updated`.

pontushojer:
Related to a question I compared how the changing the buffer size affects the script. The current default is 500 bp.

Findings:
- Runtime unaffected by buffer size in 5-50,000 bp range. 
- Low buffer size --> Low memory but less sensitive
- High buffer size --> Increased memory but more sensitive

Sensitivity in this case relates to the ability to detect and remove duplicate barcode clusters (see "Barcodes removed" in image). Thus there is some sort of tradeoff between memory and sensitivity, therefore I think 500 it still reasonable. 

image

src/blr/cli/clusterrmdup.py
FrickTobias: Ah, I see. Makes sense!

FrickTobias:
Are you sure the plot is correct? The barcodes removed are at a maximum when `BUFFER_SIZE` is the lowest, suggesting sensitivity is high (assuming they are TP).

pontushojer:
> Are you sure the plot is correct? The barcodes removed are at a maximum when `BUFFER_SIZE` is the lowest, suggesting sensitivity is high (assuming they are TP).

Some more investigation into this showed a small error in the code that caused duplicate positions to be checked against themself. This was increasingly prominent for low buffer sizes where `find_barcode_duplicates` is called more often. The error is now fixed and I can show the current data

image

Here we instead see an increase in sensitivity with increased buffer size at the expense of increased memory use. Based on this I changed the default buffer size to 200 bp which is close  just above the typical read length (150 bp). I also added buffer_size as an optional argument.

src/blr/cli/clusterrmdup.py
pontushojer: I changed the name of `orientation_ok` instead to `valid_pair_orientation`.

FrickTobias:
Awesome! Looks good!

PR #177, opened 2019-12-13
Update tagbam split

FrickTobias:
Followup PR from #165. **TL;DR:** regex was nice but slow so now it uses split instead.

Now `tagbam` runs in different modes where one can choose `--format sam` or `--format ema` which processes reads differently. 

- `--format sam` splits header on `_` where SAM tags are created from everything found after the first element. This will run any number of SAM tags found in header.

- `--format ema` rsplits once on `:` where the last element is removed. This does not set a SAM tag as it should already be in the alignment produced from EMA.

### DIFF
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -45,7 +45,7 @@ def main(args):
                         read.set_tag(tag, match.group("value"), value_type=match.group("type"))
                     summary[f"Reads with tag {tag}"] += 1
                 else:
-                    summary[f"reads without tag {tag}"] += 1
+                    summary[f"Reads without tag {tag}"] += 1
 
             out.write(read)
 
--- a/src/blr/cli/tagbam.py
+++ b/src/blr/cli/tagbam.py
@@ -30,7 +30,7 @@ def main(args):
     regex_patterns_dict = dict()
     for tag in args.tags:
         # Make regex expression, compile and save
-        pattern = regex_function(bam_tag=tag)
+        pattern = regex_function(bam_tag=tag, length=args.tag_length)
         compiled_pattern = re.compile(pattern)
         regex_patterns_dict[tag] = compiled_pattern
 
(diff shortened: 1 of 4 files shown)


src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,82 @@
...
+    :param summary: Collections's Counter object
+    :return:
+    """
 
-            if raw_barcode_tag:
-                assert raw_barcode_tag.startswith(f"{args.sequence_tag}:Z:")
-                raw_barcode = raw_barcode_tag.split(":")[-1]
-                read.set_tag(args.sequence_tag, raw_barcode, value_type='Z')
+    # Strip header
+    summary["Total reads"] += 1
```
pontushojer: I would move this statistic to the main function as it is the same independent of the function used.

src/blr/cli/tagbam.py
```diff
@@ -1,58 +1,82 @@
...
+
+from blr.utils import print_stats
 
 logger = logging.getLogger(__name__)
 
 
 def main(args):
+    # Can't be at top since function are defined later
+    FUNCTION_DICT = {
+        "sam": mode_samtags_underline_separation,
```
pontushojer: Switch to lowercase. Only constants should be upper case.

pontushojer:
I looked into the why the md5sum had changed and it has to do with the order of the RX and BX tag in the BAM file. Previously the BX tag was assigned first and then the RX tag. Now the order depends on where they appear in the header. I our case the header look like this: `ST-E00269:339:H27G2CCX2:7:1213:17898:55420_RX:Z:CGTAGTTACTCCTACGTTTG_BX:Z:CGTAGTTACTCCTACGTTTG`. Thus the RX tag will be assigned before the BX tag. 

Regarding the new md5sum for `m2` I reverted to the old check and removed the sort step as this was not needed.

FrickTobias:
Thanks for looking into it!

Regarding the travis error; seemingly travis now includes `md5sum` for osx and the install command thus causes an error when brew tries to overwrite an existing link. Removing the install command seems to have solved the problem.

Issue #179, opened 2019-12-16
Hapcut2 caclulate_haplotype_statistics only for single chromosome


pontushojer:
I got the following error when running on a full dataset:

```
ERROR: Chromosome in reference haplotype VCF doesn't match chromosome in VCF used for phasing
reference haplotype VCF: /proj/uppstore2018173/private/analysis/191128.SingleMoleculePhasing/HG002_GRCh38_GIAB_v3.3.2_highconf_triophased.vcf
chr1 != chr2
```

Reading the [docs for the tool](https://github.com/vibansal/HapCUT2/blob/master/utilities/calculate_haplotype_statistics.py) I found that the input VCF can only contain a single chromosome. 

>Calculate statistics on haplotypes assembled using HapCUT2 or similar tools.
Error rates for an assembled haplotype (specified by -v1 and optionally -h1 arguments)
are computed with respect to a "reference" haplotype (specified by -v2 and optionally -h2 arguments).
All files must contain information for one chromosome only!
To compute aggregate statistics across multiple chromosomes, provide files for
each chromosome/contig as an ordered list, using the same chromosome order between flags.
Note: Triallelic variants are supported, but variants with more than 2 alternative alleles
are currently NOT supported. These variants are ignored. Also, variants where the ref and alt
alleles differ between the test haplotype and reference haplotype are skipped.

Working on fix.

FrickTobias:
It's quite common for people to use [hap.py](https://github.com/Illumina/hap.py) for calculating phasing stats. With that said I'm not sure if they calculate the same metrics.

pontushojer:
I started testing hap.py, thank for the tip. 

However I think that this is still quite a nice statistics so I don't just want to abandon this tool. I have two possible solutions for this.

## 1) Split VCF per chromosome.
This is what i tried in #180

Pros:
- Don't need to modify and maintain calculate_haplotype_statistics script

Cons:
- Current solution quite messy 
- Requires folder structure (see discussion in #155 )
- Need to modify "ground truth" phased VCF. 


## 2) Modify calculate_haplotype_statistics to work on current VCFs.
I have made a branch using this approach called [stats-script](https://github.com/FrickTobias/BLR/tree/stats-script) which you can checkout. 

Pros:
- Don't need to modify VCF files. 

Cons:
- Need to maintain complicated script. 

I am not sure which solution is best though. Any input?

FrickTobias:
We went with solution number two, implemented here #181.

PR #180, opened 2019-12-16
Phasing stats per chromosome

pontushojer:
See issue #179 

Added handling for multiple chromosomes in phased VCF file as this cannot be handled by Hapcut2's `calculate_haplotype_statistics.py`. The proposed addition splits the VCF based on chromosomes (specified in config under `chromosomes` as a file or using the default of chr1-chr22). The final output is an CSV file with aggregated statistics for all chromosomes.

### DIFF
--- a/environment.linux.lock.yml
+++ b/environment.linux.lock.yml
@@ -9,13 +9,13 @@ dependencies:
   - blas=2.14=openblas
   - bowtie2=2.3.5=py36he860b03_0
   - bwa=0.7.17=hed695b0_6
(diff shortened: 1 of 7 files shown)


pontushojer:
This is fastly becoming an ugly solution, I think I might just modify the calculate_haplotype_statistics script from hapcut2.

pontushojer:
I am closing this PR after discussing with @FrickTobias. We will go ahead with a modified version of the calculate_phasing_stats.py script from the HapCUT2 package for now.

PR #181, opened 2020-01-14
Phasing stats script for multiple chromosomes

pontushojer:
The is a modified version of the [calculate_haplotype_statistics](https://github.com/vibansal/HapCUT2/blob/master/utilities/calculate_haplotype_statistics.py) script to solve the issue in #180.


FrickTobias:
I think will be a good addition for now. 

I'll also just write what we concluded from out discussions, that this is supposed to be a temporary solution as this isn't really a good solution in terms of maintenance.

Issue #178, opened 2019-12-16
filterclusters does not remove duplicate marked reads.


pontushojer:
`filterclusters` skips reads that are duplicate as these have not been marked with the `MN` tag required for filtering them out, i.e. removing the barcode tag. Thus they appear in the final BAM file and cause problems downstream.

PR #182, opened 2020-01-15
Filter out duplicate read in filterclusters

pontushojer:
Solves #178 

Change tagging routine in `buildmolecules` so that even duplicate read will get the number_tag (MN). This ensures that they will be filtered out correctly in the `filterclusters` step. 

Other edits:
- Refactored `buildmolecules`
- Change test config so that `filterclusters` is properly tested. 
- Updated `filterclusters` to handle duplicate reads that have the MN tag but not the MI tag.

### DIFF
--- a/src/blr/cli/buildmolecules.py
+++ b/src/blr/cli/buildmolecules.py
@@ -32,28 +32,22 @@ def main(args):
         logger.info("Writing filtered bam file")
         for read in tqdm(openin.fetch(until_eof=True)):
             name = read.query_name
+            barcode = utils.get_bamtag(pysam_read=read, tag=args.barcode_tag)
 
-            # If barcode is not in all_molecules the barcode does not have enough proximal reads to make a single
-            # molecule. If the barcode has more than <max_molecules> molecules, remove it from the read.
-            if name in header_to_mol_dict:
-                summary["Reads tagged"] += 1
-
-                # Set tags
-                molecule_ID = header_to_mol_dict[name]
-                read.set_tag(args.molecule_tag, molecule_ID)
-                bc_num_molecules = len(bc_to_mol_dict[read.get_tag(args.barcode_tag)])
+            # If barcode is not in bc_to_mol_dict the barcode does not have enough proximal reads to make a single
+            # molecule.
+            if barcode in bc_to_mol_dict:
+                summary[f"Reads {args.number_tag} tagged"] += 1
+                bc_num_molecules = len(bc_to_mol_dict[barcode])
                 read.set_tag(args.number_tag, bc_num_molecules)
 
+            # If the read name is in header_to_mol_dict then it is associated to a specific molecule.
+            if name in header_to_mol_dict:
+                summary[f"Reads {args.molecule_tag} tagged"] += 1
+                molecule_ID = header_to_mol_dict[name]
+                read.set_tag(args.molecule_tag, molecule_ID)
+
             openout.write(read)
-
-    non_analyced_keys = (
-        "Overlapping reads in molecule",
-        "Reads without barcode",
-        f"Unmapped {args.barcode_tag} tagged read",
-        "Duplicates"
-    )
-
-    summary["Non analyced reads"] = sum([summary[key] for key in non_analyced_keys])
 
     utils.print_stats(summary, name=__name__)
 
@@ -63,12 +57,35 @@ def main(args):
         write_molecule_stats(bc_to_mol_dict)
 
 
+def parse_reads(pysam_openfile, barcode_tag, summary):
+    for read in tqdm(pysam_openfile.fetch(until_eof=True)):
+        is_good = True
+        summary["Total reads"] += 1
+        if read.is_duplicate:
+            summary["Duplicates"] += 1
+            is_good = False
+
+        barcode = utils.get_bamtag(pysam_read=read, tag=barcode_tag)
+        if not barcode:
+            summary["Reads without barcode"] += 1
+            is_good = False
+
+        if read.is_unmapped and barcode:
+            summary[f"Unmapped {barcode_tag} tagged read"] += 1
+            is_good = False
+
+        if is_good:
+            yield barcode, read
+        else:
+            summary["Non analyced reads"] += 1
+
+
 def build_molecules(pysam_openfile, barcode_tag, window, min_reads, summary):
     """
     Builds all_molecules.bc_to_mol ([barcode][moleculeID] = molecule) and
     all_molecules.header_to_mol ([read_name]=mol_ID)
     :param pysam_openfile: Pysam open file instance.
-    :param barcode_tag: Tag used to store barcode in bam file (usually BC).
+    :param barcode_tag: Tag used to store barcode in bam file.
     :param window: Max distance between reads to include in the same molecule.
     :param min_reads: Minimum reads to include molecule in all_molecules.bc_to_mol
     :param summary: dict for stats collection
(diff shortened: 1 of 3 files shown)


src/blr/cli/buildmolecules.py
```diff
@@ -63,12 +57,35 @@ def main(args):
...
+            is_good = False
+
+        barcode = utils.get_bamtag(pysam_read=read, tag=barcode_tag)
+        if not barcode:
+            summary["Reads without barcode"] += 1
+            is_good = False
+
+        if read.is_unmapped and barcode:
+            summary[f"Unmapped {barcode_tag} tagged read"] += 1
+            is_good = False
```
FrickTobias: A read which isn't mapped but has a barcode will not be caught here, is there a point to that?

src/blr/cli/buildmolecules.py
```diff
@@ -32,29 +32,23 @@ def main(args):
...
+            barcode = utils.get_bamtag(pysam_read=read, tag=args.barcode_tag)
 
-            # If barcode is not in all_molecules the barcode does not have enough proximal reads to make a single
-            # molecule. If the barcode has more than <max_molecules> molecules, remove it from the read.
-            if name in header_to_mol_dict:
-                summary["Reads tagged"] += 1
+            # If barcode is not in bc_to_mol_dict the barcode does not have enough proximal reads to make a single
+            # molecule.
+            if barcode in bc_to_mol_dict:
+                summary[f"Reads {args.number_tag} tagged"] += 1
```
FrickTobias:
```suggestion
                summary[f"Output reads {args.number_tag} tagged"] += 1
```

src/blr/cli/buildmolecules.py
```diff
@@ -32,29 +32,23 @@ def main(args):
...
+                bc_num_molecules = len(bc_to_mol_dict[barcode])
+                read.set_tag(args.number_tag, bc_num_molecules)
 
-                # Set tags
+            # If the read name is in header_to_mol_dict then it is associated to a specific molecule.
+            if name in header_to_mol_dict:
+                summary[f"Reads {args.molecule_tag} tagged"] += 1
```
FrickTobias:
```suggestion
                summary[f"Output reads {args.molecule_tag} tagged"] += 1
```

src/blr/cli/filterclusters.py
```diff
@@ -54,9 +55,12 @@ def strip_barcode(pysam_read, tags_to_be_removed, removed_tags):
 
     # Remove tags
     for bam_tag in tags_to_be_removed:
-        removed_tags[bam_tag].add(pysam_read.get_tag(bam_tag))
-        # Strip read from tag
-        pysam_read.set_tag(bam_tag, None, value_type="Z")
+
+        if pysam_read.has_tag(bam_tag):
```
FrickTobias: Is this check needed? This function is only run `if no_mols and no_mols > threshold`, which shuold also mean it requires the read to have the tag right?

src/blr/cli/filterclusters.py
```diff
@@ -54,9 +55,12 @@ def strip_barcode(pysam_read, tags_to_be_removed, removed_tags):
...
+            removed_tags[bam_tag].add(pysam_read.get_tag(bam_tag))
+            # Strip read from tag
+            pysam_read.set_tag(bam_tag, None, value_type="Z")
+            summary[f"Total {bam_tag} tags removed"] += 1
```
FrickTobias: If the previous comment is true, there is no need to pass summary to this function, it can be incremented outside.

src/blr/cli/buildmolecules.py
pontushojer: This is just a different way of writing the [same code that was before](https://github.com/FrickTobias/BLR/blob/master/src/blr/cli/buildmolecules.py#L90), see below:

```
if not barcode:
            summary["Reads without barcode"] += 1
        elif read.is_unmapped:
            summary[f"Unmapped {barcode_tag} tagged read"] += 1
```

As for is this statistic is any useful I have not thought much about. Probably it is sufficient to keep track of the unmapped reads.

src/blr/cli/filterclusters.py
pontushojer: Yes it is needed. Now when duplicate reads are also to be tagged with the MN for `no_mols` they will still miss the MI tag for molecule index. Thus they would raise an error here if not checked first.

src/blr/cli/filterclusters.py
pontushojer: See my comment ;)

src/blr/cli/buildmolecules.py
FrickTobias: Ah right. But then I'd change it since we can't see the point of keeping it that way anymore.

src/blr/cli/filterclusters.py
FrickTobias: But if we add the MN tag, wouldn't it also make sense to also add the MI tag as well? 

Otherwise we'll be a bit inconsistent since duplicate reads will have some, but not all tags normal reads have. Basically I'm promoting either removing all or keeping all tags from the duplicate reads.

pontushojer: You mean adding it but with a `None` value or such? Or maybe `-1` as it has to be integers? I guess that is the alternative. 

We do have read without the BX tag (or any other of our tags) so we are not entirely consistent in that case too.

FrickTobias: I mean if a duplicate read has a `BX` tag it should also get the `MN` and `MI` tag as all other reads (which will be identical to the non-duplicate read at that position presumably). 

So instead of having 
```
read1 ... <not duplicate> BX:Z:ATGC MN:I:<num_mol> MI:<ID>
read2 ... <duplicate> BX:Z:ATGC
```
we write 
```
read1 ... <not duplicate> BX:Z:ATGC MN:I:<num_mol> MI:<ID>
read2 ... <duplicate> BX:Z:ATGC MN:I:<num_mol> MI:<ID>
```
and rely downstream analysis on not looking at duplicates if they aren't relevant. 

Essentially, for all processing I believe duplicate reads should be treated the same (such as **adding** molecule information or **merging** barcode duplicate), while excluding them from all analysis step (such as **building** the molecules or **finding** the barcode duplicates).

Does that distinguishment make sense?

pontushojer:
So I have now changed how reads are tagged in `buildmolecules`. If a read has a barcode, the MN (molecule number tag) is set globally for that barcode, otherwise it is set to **0**. If a read is associated with specific molecule the MN (molecule index tag) is set to the index of the particular molecule. Otherwise the index is set to **-1**, signifying no associated molecule. See relevant code below. 

```python
# If barcode is not in bc_to_mol_dict the barcode does not have enough proximal reads to make a single
# molecule.
if barcode in bc_to_mol_dict:
    bc_num_molecules = len(bc_to_mol_dict[barcode])
    read.set_tag(args.number_tag, bc_num_molecules)
else:
    read.set_tag(args.number_tag, 0)

# If the read name is in header_to_mol_dict then it is associated to a specific molecule.
if name in header_to_mol_dict:
    molecule_ID = header_to_mol_dict[name]
    read.set_tag(args.molecule_tag, molecule_ID)
else:
    # For reads not associated to a specific molecule the molecule id is set to -1.
    read.set_tag(args.molecule_tag, -1)
```

Also made some minor changes to summary variables.

FrickTobias:
I think this looks good!

Issue #91, opened 2019-09-16
Create a nice summary report


marcelm:
After the pipeline has done all the computations, it should create a report. We can use MultiQC for this. It may be necessary to write our own plugin for it for the custom BLR-specific parts, but we can use its existing modules already now for Bowtie2, Cutadapt, MarkDuplicates etc.

FrickTobias:
I've started working on adding some `fastqc` and `multiqc` steps to the pipe. However at this point of time I'm seeing this could be done in multiple ways which have different advantages.

1. Run `fastqc` before processing anything. This gives the user a way to check their data before running the pipe. 

2. Run `fastqc` after demultiplexing. This will require processing before running the qc but arguably a user could run an initial `fastqc` run on their raw reads if they want to. Regarding advantages this would make `fastqc` work a little better since it won't be confused by the handles. 

3. Run `fastqc` both before and after trimming. It doesn't take more processing power, the question is if this is interesting enough to include into the final output. Sometimes I think the mark of a really good program is also knowing when to not include output (provided it isn't relevant for normal usage).

Personally I'm more inclined to go for suggestion number two but I think this could be valuable to gather your input on.

pontushojer:
I have started working on a MultiQC plugin to integrate our data to the MultiQC report. See [MultiQC_BLR](https://github.com/pontushojer/MultiQC_BLR).

pontushojer:
https://github.com/NBISweden/BLR/pull/17 that integrates the BLR-plugin is now merged we have the basis for a summary report. Thus I will close this issue for now

PR #187, opened 2020-01-17
Initial report 

FrickTobias:
A start to solving issue #91. 

Changes:
- Added fastqc and multiqc
- Runs fastqc trimmed read files
- Runs multiqc at the end of the pipeline

Improvements for future:
- Rename files so the multiqc puts everything on one row (currently one per log file summarized).

### DIFF
--- a/environment.linux.lock.yml
+++ b/environment.linux.lock.yml
@@ -3,65 +3,118 @@ channels:
   - bioconda
   - defaults
 dependencies:
-  - _libgcc_mutex=0.1=main
+  - _libgcc_mutex=0.1=conda_forge
+  - _openmp_mutex=4.5=0_gnu
+  - alsa-lib=1.1.5=h516909a_1002
   - appdirs=1.4.3=py_1
   - attrs=19.3.0=py_0
   - blas=2.14=openblas
   - bowtie2=2.3.5=py36he860b03_0
-  - bwa=0.7.17=hed695b0_6
-  - bzip2=1.0.8=h516909a_1
+  - bwa=0.7.17=hed695b0_7
+  - bzip2=1.0.8=h516909a_2
   - ca-certificates=2019.11.28=hecc5488_0
   - certifi=2019.11.28=py36_0
   - cffi=1.13.2=py36h8022711_0
   - chardet=3.0.4=py36_1003
+  - click=7.0=py_0
+  - coloredlogs=10.0=py36_1000
+  - colormath=3.0.0=py_2
   - configargparse=0.13.0=py_1
-  - cryptography=2.8=py36h72c5cf5_0
+  - cryptography=2.8=py36h72c5cf5_1
   - curl=7.65.3=hf8cf82a_0
   - cutadapt=2.5=py36h516909a_0
+  - cycler=0.10.0=py_2
   - datrie=0.8=py36h516909a_0
+  - dbus=1.13.6=he372182_0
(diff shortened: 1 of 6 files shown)


environment.linux.lock.yml
```diff
@@ -3,89 +3,164 @@ channels:
...
+  - gettext=0.19.8.1=hc5be6a0_1002
+  - giflib=5.2.1=h516909a_1
   - gitdb2=2.0.6=py_0
   - gitpython=3.0.5=py_0
-  - hapcut2=1.2=hed50d52_0
-  - htslib=1.9=h244ad75_9
+  - glib=2.58.3=py36h6f030ca_1002
+  - gst-plugins-base=1.14.5=h0935bb2_0
+  - gstreamer=1.14.5=h36ae1b5_0
+  - hapcut2=1.1=hed695b0_0
```
pontushojer: Why is this downgraded?

src/blr/Snakefile
```diff
@@ -11,7 +11,48 @@ include: "rules/phasing.smk"
...
+    # Creates fastqc reports from raw read files. Output names are automatically given by fastqc.
+    output:
+        qc_1 = "trimmed.barcoded.1_fastqc.html",
+        qc_2 = "trimmed.barcoded.2_fastqc.html",
+        zip_1 = "trimmed.barcoded.1_fastqc.zip",
+        zip_2 = "trimmed.barcoded.2_fastqc.zip"
+    input:
+        read_1 = "trimmed.barcoded.1.fastq.gz",
+        read_2 = "trimmed.barcoded.2.fastq.gz"
+    log:
```
pontushojer: You can use wildcard expansion to reduce the number of argument.

For example
```
qc_1 = "trimmed.barcoded.1_fastqc.html",
qc_2 = "trimmed.barcoded.2_fastqc.html"
```

can become
```
qc = "trimmed.barcoded.{read_nr}_fastqc.html"
```
which will run for read1 and read2 separately.


Or alternatively you can use expansion to run them together.
```
qc = expand("trimmed.barcoded.{read_nr}_fastqc.html", read_nr=(1, 2))
```

tests/run.sh
```diff
@@ -32,7 +32,7 @@ m=$(samtools view mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam | md5sum | cut -f
 test $m == 60b2acc02c872c30d8683bd6a9d3568b
 
 # Test phasing
-blr run phasing_stats.txt
+blr run
```
pontushojer: This can be removed as `blr run` is already executed within this test script.

environment.linux.lock.yml
FrickTobias: For some reason it downgraded when I added multiqc and fastqc, not quite sure why. I'll see if running the condalock upgrades it again.

FrickTobias: So seemingly conda downgrades hapcut2 from 1.2 to 1.1 when these programs are in the environment files, and I don't know why. Possibly it could have to do with

I solved it by specifying version to 1.2 in the main environment file, which seems to work.

pontushojer: Strange. Good that specifying to 1.2 worked.

FrickTobias:
PS. With the conflict merge the md5sum will probably need to be changed, waiting for Travis.

pontushojer:
Rebased PR and fixed extra parenthesis in filterclusters causing error.

Issue #186, opened 2020-01-17
Unknown target names


FrickTobias:
In `blr run` we're currently referring to targets, which can specified to run only specific parts of the pipeline. However, currently there is no way of accessing what possible targets there are in the pipeline without opening the snakemake file and reading the file.

Not only is this unaccessible to users but I've found myself needing to open the file (possibly also needing to close and then open another rule file), copy the string I'm interested in and then running the command I want to run. 

**TL;DR:** I't hard to know what possible values you can give as targets when running `blr run`.

FrickTobias:
I've implemented a possible quick and dirty solution to this which relies on having the docstrings of the rules defined the outputs for the rules. I'll list some flaws I see and I'll gladly take any feedback, advice or if anyone has another solution. 

Flaws: 
- The docstrings need to be maintained manually as it's written now 
- It is a very long/hard to read help message
- It has to be run in a folder with a blr.yaml file

See branch [sm-targets](https://github.com/FrickTobias/BLR/tree/sm-targets).

```
$ blr run -l
```
```
SETTINGS FOR: run dryrun: False
 cores: 4
 keepgoing: False
 unlock: False
 dag: False
 listrules: True
 targets: []
trim_and_tag
    
    output:
        r1_fastq="trimmed.barcoded.1.fastq.gz",
        r2_fastq="trimmed.barcoded.2.fastq.gz"
    
extract_DBS
    
    output:
        fastq="barcodes.fasta.gz"
    
hapcut2_extracthairs
    
    output:
        unlinked = "mapped.sorted.tag.mkdup.bcmerge.mol.filt.unlinked.txt"
    
hapcut2_linkfragments
    
    output:
        linked = "mapped.sorted.tag.mkdup.bcmerge.mol.filt.linked.txt"
    
hapcut2_phasing
    
    output:
        phase = "mapped.sorted.tag.mkdup.bcmerge.mol.filt.phase",
        phased_vcf = "mapped.sorted.tag.mkdup.bcmerge.mol.filt.phase.phased.VCF"
    
hapcut2_stats
    
    output:
        stats = "phasing_stats.txt"
    
final
    
    blr run default targets:
        "reads.1.final.fastq.gz",
        "reads.2.final.fastq.gz",
        "phasing_stats.txt",
        "multiqc_report.html"
    
help
    
    Print list of all targets with help.
    
fastqc_raw_reads
    
    output:
        qc_1 = "reads.1_fastqc.html",
[...]

PR #188, opened 2020-01-20
Add filegraph to run command.

pontushojer:
Solves #186.

Filegraph can show input and output file for the rules in the pipeline.


FrickTobias:
Awesome!

PR #185, opened 2020-01-17
Add processing information to BAM header.

pontushojer:
I implemented a utility function that can add processing information to the BAM header. Currently this will add information about the program used (blr), the tool used (tagbam, clusterrmdup, etc.) and the command line .. command? (e.g. what was executed). Below is an example with the header from the final `*filt.bam` file. Note that the order of execution of each command is also maintained. What I would like to add later on is some kind of version number for the program. 

This addition will make it easy to track the exact processing performed on any BAM file, even should it be renamed.  

**Example. samtools view -H mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam**
```  
@HD	VN:1.0	SO:coordinate
@SQ	SN:chr1mini	LN:100001
@PG	PN:bowtie2	ID:bowtie2	CL:"/Users/pontus.hojer/miniconda3/envs/blr/bin/bowtie2-align-s --wrapper basic-0 -p 4 --reorder --maxins 2000 -x ../testdata/chr1mini.fasta -1 trimmed.barcoded.1.fastq.gz -2 trimmed.barcoded.2.fastq.gz"	VN:2.3.5
@PG	PN:blr	ID:tagbam	CL:"/Users/pontus.hojer/miniconda3/envs/blr/bin/blr tagbam mapped.sorted.bam -o mapped.sorted.tag.bam -f sam"
@PG	ID:sambamba	CL:markdup -t 4 mapped.sorted.tag.bam mapped.sorted.tag.mkdup.bam	PP:tagbam	VN:0.6.6
@PG	PN:blr	ID:clusterrmdup	CL:"/Users/pontus.hojer/miniconda3/envs/blr/bin/blr clusterrmdup mapped.sorted.tag.mkdup.bam barcode-merges.csv -o mapped.sorted.tag.mkdup.bcmerge.bam -b BX"
@PG	PN:blr	ID:buildmolecules	CL:"/Users/pontus.hojer/miniconda3/envs/blr/bin/blr buildmolecules mapped.sorted.tag.mkdup.bcmerge.bam -o mapped.sorted.tag.mkdup.bcmerge.mol.bam --stats-files -m MI -n MN -b BX"
@PG	PN:blr	ID:filterclusters	CL:"/Users/pontus.hojer/miniconda3/envs/blr/bin/blr filterclusters mapped.sorted.tag.mkdup.bcmerge.mol.bam -m MI -n MN -b BX -M 260"
```

PR #189, opened 2020-01-21
Fix streaming issue from PR#185

pontushojer:
PR #185 modifies the headers of output SAM/BAM files with processing information. This was done separately from the reading of the file in the processing script so the file would have to be read twice in order to get the new header and read the file. This is not great if we would like to stream files. Thus I have here added a combined reader/writer class that enables streaming.

### DIFF
--- a/src/blr/utils.py
+++ b/src/blr/utils.py
@@ -103,9 +103,9 @@ class PySAMIO:
 
     def _make_header(self, name):
         """
-        Create SAM header dict with new tool and command line argument information based on template file. Appends new PG
-        entry with tool name (ID), software name (PN), command line arguments (CL) to track the tools applied to the file.
-        Use in output SAM/BAM file as 'header' attribute.
+        Create SAM header dict with new tool and command line argument information based on template file. Appends
+        new PG entry with tool name (ID), software name (PN), command line arguments (CL) to track the tools applied
+        to the file. Use in output SAM/BAM file as 'header' attribute.
 
         :param name: string. Pass '__name__' variable to be used to get program and tool name.
         :return: pysam.AlignmentHeader object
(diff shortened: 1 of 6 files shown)


src/blr/utils.py
```diff
@@ -79,49 +79,68 @@ def print_stats(summary, name=None, value_width=15, print_to=sys.stderr):
...
         """
-        Check that identifier does not appear in previous header entries. Update with numbered suffix if the name is
-        not unique.
-        :param identifier: string. Program record identifier
-        :param prev_entries: dict. Dictionary of "PG" header entries from pysam.AlignmentFile header.
-        :return: string. Updated identifier
+        :param inname: Path to input SAM/BAM file.
+        :param inmode: Reading mode for input file. 'r' for SAM and 'rb' for BAM.
+        :param outname: Path to output SAM/BAM file.
+        :param outmode: Reading mode for output file. 'r' for SAM and 'rb' for BAM.
```
FrickTobias: If we're making a class for doing everything automatically I think we could go for automatically detecting format from the file name, otherwise defaulting to BAM (as I see it most commonly used).

src/blr/utils.py
```diff
@@ -79,49 +79,68 @@ def print_stats(summary, name=None, value_width=15, print_to=sys.stderr):
...
+        self.outfile = pysam.AlignmentFile(outname, outmode, header=self.header)
 
-        # Check that id is unique and does not appear in header, otherwise add number at end.
-        id_name = check_name(id_name, pg_entries)
+    def __enter__(self):
+        return self.infile, self.outfile
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.infile.close(), self.outfile.close()
+        return isinstance(exc_val, OSError)
```
FrickTobias: \*Reading up about double leading/trailing\* - Cool!

pontushojer:
@FrickTobias Made BAM format default in all programs for now.

Issue #184, opened 2020-01-16
Add multiple SNV callers


FrickTobias:
Currently there is only one choice of SNV caller. It would be beneficial to investigate if multiple ones can be implemented, investigated an offered as alternatives when running the pipeline. Below are some well-regarded SNV callers

Implemented SNV callers:
- [x] GATK
- [x] SAMtools
- [x] FreeBayes
- [x] DeepVariant*

*DeepVariant is only supported on Linux-based unix type system so it will not be available on macOS. However from reading their paper I think it is worth looking into if it can implemented on compatible systems.

FrickTobias:
Now GATK and SAMtools have been added in the [var-callers](https://github.com/FrickTobias/BLR/tree/var-callers) branch, now proceeding to test some performance. 

I also still need to look into the outputs, specifically because gatk isn't exactly the same meaning it might cause downstream problems.

I'll also include the code necessary to run DeepVariant (but not mentioned in the help message), which I've tested on UPPMAX but I'm not currently able to test setting that up as I'm not running a compatible system and I don't have sudo privileges at UPPMAX.

FrickTobias:
## Results

So here are some results from the variant calling. Although after running and seeing the results we should probably also run a more thorough benchmarking for different coverages. In the case we have here we see `gatk` performing really well, but that might change for when coverages are of more like the sequencing depths one might expect from a whole genome sequencing setup.

All results where compared using `vcftools --diff-site` with a minimum phred score of 50 and the original read file was 11x coverage.

**Result highlights**:
- `gatk` performs really well (assuming variants are filtered)
- `deepvariant` possibly has a more stringent quality assignment
- `deepvariant` possibly does not work with low sequencing depths
- preliminary results indicate `freebayes` and `bcftools` are not as good as `gatk`
- This needs to be investigated further to understand under what conditions the softwares work well

### **chr22, min qual 50** 
**NB:** I just noticed I switched the metric/score rows for the raw number of calls, but I don't want to redo it in markdown since it's a bit tedious.

Metric | Score
|----|----|
**bcftools**
sensitivity | 0.7310736658
accuracy estimate | 0.9310451453
"confirmed accuracy" | 0.5323738463
30110| union
8846| only_ref
24218| only_calls
2230| conflicting
**deepvariant** 
sensitivity | 0.1269606177
accuracy estimate | 0.9990447077
"confirmed accuracy" | 0.785253041
5229| union
35952| only_ref
1425| only_calls
[...]

PR #191, opened 2020-01-23
Add multiple variant callers

FrickTobias:
Adding these to the pipe for variant calling:
- gatk
- bcftools

Some preliminary benchmarking results can be seen in #184.

This also adds the code to run DeepVariant but I haven't implemented a workaround for only installing it on Linux, but it is available on conda.


FrickTobias:
*Some merge conflicts later...

src/blr/Snakefile
```diff
@@ -247,15 +247,73 @@ rule symlink_reference_variants:
         "ln -s {config[reference_variants]} {output}"
 
 
-rule call_variants_freebayes:
+rule add_RG_tag:
```
pontushojer: The RG tag is only needed for the GATK software right? Do you know if it affects the other variant callers? If it does not affect them i would suggest to just incorporate all these edits into the `tagbam` script. That way we don't have to read the file one more time. 

This also becomes handy if we want to merge multiple runs after mapping, such as for the SM libs I run before christmas.

pontushojer: Alternatively you could also use `samtools addreplacerg`.

Something like this should work.

```
samtools addreplacerg -r ID:1 -r LB:lib1 -r PU:unit1 -r SM:20 -r PL:ILLUMINA -O BAM in.bam > out.bam
```

src/blr/Snakefile
```diff
@@ -247,15 +247,73 @@ rule symlink_reference_variants:
...
+                " --ploidy GRCh38"
+                " --threads {threads}"
+                " -o {output.vcf} 2>> {log}",
+            "gatk":
+                "gatk --java-options -Xmx{config[heap_space]}g HaplotypeCaller"
+                " -R {config[genome_reference]}"
+                " -I {input.bam}"
+                " -ERC GVCF"
+                " -O {output.vcf} 2> {log}",
+            "deepvariant":
```
pontushojer: Maybe add a comment here in the code or a TODO that this is not implemented.

pontushojer:
Looks mostly good!

I had a comment about how the RG tagging is implemented which I think we should change if possible.

src/blr/Snakefile
FrickTobias: Went with samtools since it involves less maintenance.

src/blr/config.schema.yaml
```diff
@@ -52,4 +52,9 @@ properties:
   max_molecules_per_bc:
     type: integer
     description: Max number of molecules allowed for a single barcode (removes if bc has > 260 molecules).
-
+  variant_caller:
+    type: string
```
pontushojer:
```suggestion
    type: string
    pattern: "gatk|bcftools|freebayes|deepvariant"
```
Add a pattern for allowed inputs.

src/blr/Snakefile
```diff
@@ -247,15 +247,80 @@ rule symlink_reference_variants:
...
-    log: "call_variants_freebayes.log"
+        bam = "mapped.sorted.tag.mkdup.bcmerge.mol.filt.bam"
+    log: "add_RG_tag.log"
     shell:
-         "freebayes"
-         " -f {config[genome_reference]}"
-         " {input.bam} |"
-         "vcffilter -f 'AC > 0' 1>"
-         " {output.vcf} 2> {log}"
+        "samtools addreplacerg"
```
pontushojer: This is good but i still think we should move this to another rule to limit read/writing of files. My suggestion would be to stream the output from `filterclusters` through this command, something like...
```
blr filterclusters input.bam ... | samtools addreplacerg - ... | tee out.bam | samtools index. 
```
This also means one less step of indexing.

FrickTobias: Yeah that's right, forgot about that. Makes total since this should more or less be only read/write and if it ever comes close to be CPU throttling we can add threading to it. 

I added it even further up in the tagbam rule, thinking that seemed like a good rule name for incorporation of any number of SAM tags. 

Possibly it even makes sense to pipe this directly from the mapper (since at least broad institute software assume you set RG tags for all libraries). However, I opted not to do that for now since it's already quite a large piping operation, thinking this could is more readable.

Issue #192, opened 2020-01-27
Incorporate BQSR


FrickTobias:
It seems common to use BQSR from gatk before calling variants and it is recommended in best practises by gatk so adding it to the pipeline would probably be beneficial.

However, there seems to be cases where it [shouldn't be applied ](https://www.biostars.org/p/280274/)(or at least there is no strict consensus on the subject) so it should probably be optional.

PR #193, opened 2020-01-27
Add option to run BQSR before variant calling

FrickTobias:
Solves #192.

- Adds BQSR
- Adds dbSNP file to testfiles (v0.2.1)
- md5sum changes because of BQSR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ before_install:
   - conda update -q conda
   - conda config --add channels bioconda --add channels conda-forge
   - conda info -a
-  - wget https://export.uppmax.uu.se/uppstore2018173/blr-testdata-0.2.tar.gz
-  - tar xf blr-testdata-0.2.tar.gz
-  - ln -s blr-testdata-0.2 testdata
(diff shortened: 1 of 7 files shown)


pontushojer:
One comment about implementation.

Currently this is run for whichever variant caller is used. For `freebayes` this is not necessary (see [best practices](https://github.com/ekg/freebayes#best-practices-and-design-philosophy). I don't know how this is for bcftools or deepvariant but we should not run it unless it adds a benefit. In this case i think we should limit it to run only when using `gatk` as variant caller.

FrickTobias:
I think it's worth keeping having the possibility to decide for yourself, and rather give one recommended pipeline and possibly some recommended related changes for whenever deviations from the recommendation is made. Like if using `freebayes`, also consider changing `BQSR` to `false` or whatever we might find.

So I'd suggest to change the documentation (aka remove recommended to recommended for `gatk`) and possibly changing default of `BQSR` to `false` until we've investigated the final pipeline for optimal performance.

This way, someone who has identified they should either use or not use BQSR because of their library can easily change that without being forced to switch variant caller.

pontushojer:
That is true. Lets keep it optional for now and then we can include or recommendations in the docs. I will go ahead and merge.

Issue #50, opened 2019-06-28
Add new option to filterclusters script to include Tn5 related overlaps.


pontushojer:
Related to [comment](https://github.com/FrickTobias/BLR/commit/b9119f19c13b126cc28c67d1c01a406f9fc5de69#r34116604)

This is ment as a future improvement and is no current issue. Just writing it here to remember.

FrickTobias:
FYI: This is being investigated (and is almost done by @Parharn).

PR #201, opened 2020-02-03
Handle Tn5-overlaps in buildmolecules. 

pontushojer:
Added ability to include Tn5 type overlaps for blr libraries. see issue #50. Tn5 overlap defined as being between 8-10 bp i.e. (9+/-1bp). In the test dataset this incompased 320/520 overlapps. Also updated RG-tag info with the `library_type` from the config file.


FrickTobias:
Nice! I'll wait with merging until prior PR are merged!

